PR #1, opened 2022-02-07
Initialize the repositry with operator-sdk

arjunrn:
This change was produced by:

1. Initializing the repositry with `operator-sdk`.
   ```bash
   operator-sdk init --domain openshift.io --project-name aws-load-balancer-operator \
   --repo github.com/openshift/aws-load-balancer-operator
   ```
2. Creating a custom resource of type `AWSLoadBalancerController`.
   ```bash
   operator-sdk create api --group networking.olm --kind AWSLoadBalancerController \
   --namespaced=false --version v1alpha1
   ```
3. Vendor the dependencies `go mod vendor`.
4. Switch all targets in the _Makefile_ to use the vendored
   dedependencies.
5. Change the operator image to use UBI.


api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "networking.olm.openshift.io", Version: "v1alpha1"}
```
alebedev87, line 29: Doesn't `networking` look too general for this operator? Like it it was part of a bigger networking OLM API.

bundle/tests/scorecard/config.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: scorecard.operatorframework.io/v1alpha3
+kind: Configuration
```
alebedev87, line 2: Will we try to use the scorecard tests for this operator?

api/v1alpha1/groupversion_info.go
arjunrn: What do you recommend?

arjunrn: The `IngressController` belongs to the group `config.openshift.io`. My thinking here was the types are part of the `networking` group which is distributed through `olm` on `openshift.io`.

bundle/tests/scorecard/config.yaml
arjunrn: I just committed all the stuff which was generated because I wasn't sure if it would break anything else which is generated later using the `operator-sdk` command. I suppose we can delete these tests.

api/v1alpha1/groupversion_info.go
alebedev87: > What do you recommend?

Hm, good question.
I don't see a lot of examples in `olm.openshift.io` domain, only our ExternalDNS which is `externaldns.olm.openshift.io`. MetalLB Operator is in its own group: `metlalb.io`.

Something like `aws-lb-controller.networking.olm.openshift.io` indeed looks not as redundant as `externaldnses.externaldns.olm.openshift.io`.

I'm just wondering how such a group can be coordinated if tomorrow we add Azure Gateway Ingress Controller, for instance. Will we have to have to upgrade the APIs of all the controllers to a new version (e.g. `v1alpha1` -> `v1beta1`) if just one of them changes?

arjunrn: > Something like aws-lb-controller.networking.olm.openshift.io indeed looks not as redundant as externaldnses.externaldns.olm.openshift.io

That is exactly what it is now. Because the full name will be `awsloadbalancercontroller.networking.olm.openshift.io`. Just the group name is `networking.olm.openshift.io`.

arjunrn: > I'm just wondering how such a group can be coordinated if tomorrow we add Azure Gateway Ingress Controller, for instance. Will we have to have to upgrade the APIs of all the controllers to a new version (e.g. v1alpha1 -> v1beta1) if just one of them changes?

No each one will be upgraded separately.

Makefile
```diff
@@ -0,0 +1,225 @@
...
+# For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
+# openshift.io/aws-load-balancer-operator-bundle:$VERSION and openshift.io/aws-load-balancer-operator-catalog:$VERSION.
+IMAGE_TAG_BASE ?= openshift.io/aws-load-balancer-operator
+
+# BUNDLE_IMG defines the image:tag used for the bundle.
+# You can use it as an arg. (E.g make bundle-build BUNDLE_IMG=<some-registry>/<project-name-bundle>:<tag>)
+BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+
+# Image URL to use all building/pushing image targets
+IMG ?= controller:latest
```
bharath-b-rh, line 39: Would it be better if IMG variable is assigned $(IMAGE_TAG_BASE):v$(VERSION) as image URL?

PR #8, opened 2022-02-21
Initial controller spec generated

thejasn:
Ref: CFE-313


api/v1alpha1/awsloadbalancercontroller_types.go
```diff
@@ -20,22 +20,85 @@ import (
...
 
-	// Foo is an example field of AWSLoadBalancerController. Edit awsloadbalancercontroller_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// SubnetTagging describes how resource tagging will be done by the operator.
+	//
+	// When in "Auto", the operator will detect the subnets where the load balancers
+	// will be provisioned and have the required resource tags on them. Whereas when
+	// set to manual, this responsibility lies on the user.
+	//
+	SubnetTagging SubnetTaggingPolicy `json:"subnetTagging,omitempty"`
```
alebedev87, line 53: `omitempty` implies that this is an optional field, let's add the optional comment markers then.

However don't we want to set a different defaulting strategy for this field?
Let's say we can default to `ManualSubnetTaggingPolicy`. This would make the API definition slightly more approachable for the end users.

api/v1alpha1/awsloadbalancercontroller_types.go
```diff
@@ -20,22 +20,85 @@ import (
...
+	//
+	// +kubebuilder:validation:Optional
+	// +optional
+	IngressClass string `json:"ingressClass,omitempty"`
+
+	// Config specifies further customization options for the controller's deployment spec.
+	//
+	// +kubebuilder:validation:Optional
+	// +optional
+	Config AWSLoadBalancerDeploymentConfig `json:"config,omitempty"`
```
alebedev87, line 76:
```suggestion
	Config *AWSLoadBalancerDeploymentConfig `json:"config,omitempty"`
```
To not encode empty `AWSLoadBalancerDeploymentConfig` structure.

api/v1alpha1/awsloadbalancercontroller_types.go
```diff
@@ -20,22 +20,85 @@ import (
...
+	//
+	// This value is required so that this controller can function as expected
+	// in parallel to openshift-router.
+	//
+	// +kubebuilder:validation:Optional
+	// +optional
+	AdditionalResourceTags map[string]string `json:"additionalResourceTags,omitempty"`
+
+	// IngressClass specifies the Ingress class which the controller will reconcile.
+	// This will default to "alb".
```
alebedev87, line 66: I don't see the corresponding comment marker about this defaulting.

api/v1alpha1/awsloadbalancercontroller_types.go
arjunrn, line 53: This should default to `Auto`.

api/v1alpha1/awsloadbalancercontroller_types.go
```diff
@@ -20,22 +20,85 @@ import (
...
+	EnabledAddons []AWSAddon `json:"enabledAddons,omitempty"` // indicates which AWS addons should be disabled.
+
 }
 
-// AWSLoadBalancerControllerStatus defines the observed state of AWSLoadBalancerController
+type AWSLoadBalancerDeploymentConfig struct {
+
+	// +kubebuilder:validation:Optional
+	// +optional
+	Replicas int `json:"replicas,omitempty"`
```
alebedev87, line 91: So, it will be literally not possible to set `replicas` to `0` as it's the default value that will be omitted.

Also, not a big deal but since `AWSLoadBalancerControllerStatus.ObservedGeneration` is set to `int64`, I propose to set to a concrete `int` type this field too: `int32`. This would be more in the way of the API guidelines.

api/v1alpha1/awsloadbalancercontroller_types.go
```diff
@@ -20,22 +20,85 @@ import (
...
+}
+
+// AWSLoadBalancerControllerStatus defines the observed state of AWSLoadBalancerController.
 type AWSLoadBalancerControllerStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Conditions is a list of operator-specific conditions
+	// and their status.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
alebedev87, line 98: Just to be consistent, let's add optional comment markers for `Conditions` and `ObservedGeneration`.

api/v1alpha1/awsloadbalancercontroller_types.go
thejasn: Added a default value to replicas as 2 and fixed data type.

api/v1alpha1/awsloadbalancercontroller_types.go
thejasn: fixed

api/v1alpha1/awsloadbalancercontroller_types.go
thejasn: fixed

api/v1alpha1/awsloadbalancercontroller_types.go
thejasn: Added

api/v1alpha1/awsloadbalancercontroller_types.go
thejasn: Added

thejasn: Added

api/v1alpha1/awsloadbalancercontroller_types.go
alebedev87: Cosmetic comment.
Sometimes we use `+optional` marker, sometimes both `+kubebuilder:validation:Optional` and `+optional`.
Still not quite sure whether we need to have both markers but just to keep things consistent: would you mind to align all the markers, @thejasn?

alebedev87: Asked the question in kubebuilder channel of Kubernetes Slack. Both markers have the same effect in kubebuilder. `+kubebuilder:validation` seems to be kubebuilder specific while `+optional` is defined in [Kube API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md). So if the API is expected to be used by other code generators ([example](https://github.com/kubernetes/code-generator)): better to keep both markers. If only kubebuilder is expected to be used: `+kubebuilder:validation` is enough.

arjunrn: I just realized after approving this PR that the markers are not really needed because the struct is not marked as required. So I've removed them [here](https://github.com/openshift/aws-load-balancer-operator/pull/9).

PR #9, opened 2022-02-23
Remove redundant Optional markers

arjunrn:
The optional markers are only required if the parent is marked as required.


alebedev87:
I saw that the default mode was changing in the past from required to optional. I think that it is a common convention to be precise about each field, Kube API conventions around [optional vs required](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) state that optional fields have to have `+optional` marker.

arjunrn:
@alebedev87 Yes, considering the Kube API conventions we should probably keep them.

PR #10, opened 2022-02-24
Add reconciliation for Credential Requests

thejasn:
Ref: [CFE-314](https://issues.redhat.com/browse/CFE-314)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,23 @@ CONTAINER_ENGINE ?= docker
 OPERATOR_SDK_VERSION = v1.17.0
 
 GOLANGCI_LINT ?= go run github.com/golangci/golangci-lint/cmd/golangci-lint
+## iamctl vars
+
+# Assets folder for iamctl cli.
+IAMCTL_ASSETS_DIR ?= ./assets
+
+# Output path for generated file(s).
+IAMCTL_OUTPUT_DIR ?= ./pkg/controllers/awsloadbalancercontroller
+
+# Generated file name.
+IAMCTL_OUTPUT_FILE ?= iam_policy.go
+
+# Go Package of the generated file.
+IAMCTL_GO_PACKAGE ?= awsloadbalancercontroller
+
+# Built go binary path.
+IAMCTL_BINARY ?= ./bin/iamctl
+
 
 .PHONY: all
 all: build
@@ -92,7 +109,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 	hack/sync-upstream-crds.sh
 
 .PHONY: generate
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
+generate: controller-gen iamctl-gen ## Generate code containing DeepCopy, DeepCopyInto, DeepCopyObject method implementations and iamctl policies.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 .PHONY: fmt
@@ -102,6 +119,12 @@ fmt: ## Run go fmt against code.
 .PHONY: vet
 vet: ## Run go vet against code.
 	go vet -mod=vendor ./...
+
+.PHONY: iamctl-gen
+iamctl-gen: iamctl-build
+	$(IAMCTL_BINARY) -i $(IAMCTL_ASSETS_DIR)/iam-policy.json -o $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE) -p $(IAMCTL_GO_PACKAGE)
+	go fmt -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
+	go vet -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
 
 ENVTEST_ASSETS_DIR ?= $(shell pwd)/bin
 
@@ -127,6 +150,11 @@ image-build: test ## Build container image with the manager.
 .PHONY: image-push
 image-push: ## Push container image with the manager.
 	${CONTAINER_ENGINE} push ${IMG}
+
+.PHONY: iamctl-build
+iamctl-build: fmt vet ## Build iamctl binary.
+	cd ./cmd/iamctl && go build -mod=vendor -o $(IAMCTL_BINARY) . 
+	mv ./cmd/iamctl/$(IAMCTL_BINARY) ./bin/
 
 ##@ Deployment
 
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -91,6 +91,18 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - serviceaccounts
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
+          - ""
+          resources:
           - services
           verbs:
           - create
@@ -140,6 +152,14 @@ spec:
           - update
           - watch
         - apiGroups:
+          - config.openshift.io
+          resources:
+          - infrastructures
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
           - networking.olm.openshift.io
           resources:
           - awsloadbalancercontrollers
@@ -166,7 +186,7 @@ spec:
           - patch
           - update
         - apiGroups:
-          - rbac.k8s.io
+          - rbac.authorization.k8s.io
           resources:
           - clusterrolebindings
           - clusterroles
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -91,18 +91,6 @@ spec:
         - apiGroups:
           - ""
           resources:
-          - serviceaccounts
-          verbs:
-          - create
-          - delete
-          - get
-          - list
-          - patch
-          - update
-          - watch
-        - apiGroups:
-          - ""
-          resources:
           - services
           verbs:
           - create
--- a/main.go
+++ b/main.go
@@ -54,6 +54,8 @@ func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
 	utilruntime.Must(networkingolmv1alpha1.AddToScheme(scheme))
+
+	utilruntime.Must(cco.AddToScheme(scheme))
 	//+kubebuilder:scaffold:scheme
 
 	utilruntime.Must(configv1.Install(scheme))
--- a/pkg/controllers/awsloadbalancercontroller/controller.go
+++ b/pkg/controllers/awsloadbalancercontroller/controller.go
@@ -42,8 +42,8 @@ import (
 
 const (
 	controllerName               = "cluster"
-	controllerSecretName         = "aws-load-balancer-operator-credentials"
-	controllerServiceAccountName = "aws-load-balancer-operator-controller-manager"
+	controllerSecretName         = "albo-cluster-credentials"
+	controllerServiceAccountName = "albo-cluster-sa"
 )
 
 // AWSLoadBalancerControllerReconciler reconciles a AWSLoadBalancerController object
@@ -59,7 +59,6 @@ type AWSLoadBalancerControllerReconciler struct {
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/finalizers,verbs=update
 //+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="config.openshift.io",resources=infrastructures,verbs=get;list;watch
 //+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings;clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
--- a/pkg/controllers/awsloadbalancercontroller/controller.go
+++ b/pkg/controllers/awsloadbalancercontroller/controller.go
@@ -21,13 +21,15 @@ import (
 	"fmt"
 	"strings"
 
-	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
 	arv1 "k8s.io/api/admissionregistration/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+
+	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -38,7 +40,11 @@ import (
 	"github.com/openshift/aws-load-balancer-operator/pkg/aws"
 )
 
-const allowedResourceName = "cluster"
+const (
+	controllerName               = "cluster"
+	controllerSecretName         = "aws-load-balancer-operator-credentials"
+	controllerServiceAccountName = "aws-load-balancer-operator-controller-manager"
+)
 
 // AWSLoadBalancerControllerReconciler reconciles a AWSLoadBalancerController object
 type AWSLoadBalancerControllerReconciler struct {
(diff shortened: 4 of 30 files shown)


Makefile
```diff
@@ -62,6 +62,24 @@ CONTAINER_ENGINE ?= docker
 
 OPERATOR_SDK_VERSION = v1.17.0
 
+## iamctl vars
+
+# Assets folder for iamctl cli.
+ASSETS_DIR ?= ./assets
+
+# Output path for generated file(s).
+OUTPUT_DIR ?= ./pkg/controllers/awsloadbalancercontroller
```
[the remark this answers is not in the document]
thejasn: prefix added. The `iamctl-gen` is just the Makefile target, I didn't to merge it with the existing Makefile target `generate`.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -33,12 +35,13 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
-
-	albo "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
-	"github.com/openshift/aws-load-balancer-operator/pkg/aws"
 )
 
-const allowedResourceName = "cluster"
+const (
+	controllerName               = "cluster"
```
thejasn, line 41: @arjunrn I know the controller name was decided to be `cluster` in the EP, but is there any reason why we decided on it being called cluster(other than it denoting a single instance across the cluster)?  Because naming other related resources gets kind of weird (sa to be called `cluster-sa`).

Makefile
```diff
@@ -101,6 +119,12 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet -mod=vendor ./...
 
+.PHONY: iamctl-gen
+iamctl-gen: iamctl-build
+	$(IAMCTL_BINARY) genpolicy -i $(IAMCTL_ASSETS_DIR)/iam-policy.json -o $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE) -p $(IAMCTL_GO_PACKAGE)
+	go fmt -mod=vendor ./...
```
arjunrn, line 125: Why is is necessary to run `fmt` and `vet` on all the packages after generating the policy.

pkg/controllers/awsloadbalancercontroller/controller.go
arjunrn: Not necessarily. Since names of different types can be the same. You can give them the same name. So the name of the service account would be `albo-cluster`. It can be some common prefix(albo) followed by the parent resource name i.e. `cluster.

arjunrn (commented):
The PR looks good. I would prefer to see all the `iamctl` source files at the same level in a single directory. We don't need multiple folders because it's relatively small. Also there are no tests for `iamctl`. Would you like to add them in a followup PR?

cmd/iamctl/aws/ast.go
```diff
@@ -0,0 +1,213 @@
...
+			Elts: exprs,
+		}
+	default:
+		panic("unsported type for key/val expr")
+	}
+}
+
+func GenerateIAMPolicy(input, output, pkg string) {
+
+	in := strings.ReplaceAll(filetemplate, "main", pkg)
```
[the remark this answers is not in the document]
thejasn: done

cmd/iamctl/aws/ast.go
```diff
@@ -0,0 +1,213 @@
...
+	policy := iamPolicy{}
+
+	err = json.Unmarshal([]byte(jsFs), &policy)
+	if err != nil {
+		panic(fmt.Errorf("failed to parse policy JSON %v", err))
+	}
+
+	// Minifying here as a workaround for current limitations
+	// in credential requests lenght (2048 max bytes).
+	miniPoliicy := minify(policy)
```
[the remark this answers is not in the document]
thejasn: done

cmd/iamctl/aws/ast.go
```diff
@@ -0,0 +1,213 @@
...
+
+	exprs := make([]ast.Expr, 0, len(miniPoliicy.Statement))
```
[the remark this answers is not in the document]
thejasn: done

cmd/iamctl/aws/ast.go
```diff
@@ -0,0 +1,213 @@
...
+			Value: buildStrings(p.Resource[0]),
+		}
+
+		policyList[3] = &ast.KeyValueExpr{
+			Key:   ast.NewIdent(policycondition),
+			Value: buildKeyValueExpr(p.Condition),
+		}
+
+		exprs = append(exprs, &ast.CompositeLit{Elts: policyList})
+	}
```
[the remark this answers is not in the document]
thejasn: done

cmd/iamctl/aws/minify.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// Minify replaces strict actions allowed across resources with much weaker
+// actions using wildcards(*). It also removes any resource and condition limits
+// applied in the policy. This reduces policy granularity and relies on
+// the operator for adhering to resource access. NOTE: Potential security concern.
+func minify(policy iamPolicy) iamPolicy {
+	var miniPolicy iamPolicy
+
+	resources := make(map[string]int)
```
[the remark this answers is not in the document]
thejasn: done

main.go
```diff
@@ -20,27 +20,25 @@ import (
 	"context"
 	"flag"
 	"os"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	configv1 "github.com/openshift/api/config/v1"
```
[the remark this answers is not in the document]
thejasn: done

Makefile
thejasn: done

cmd/iamctl/aws/ast.go
```diff
@@ -0,0 +1,213 @@
...
+
+	astutil.Apply(file, nil, func(c *astutil.Cursor) bool {
```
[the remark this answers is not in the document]
thejasn: It doesn't provide any direct way to format generated code. It exposes internal using [Pos](https://pkg.go.dev/go/token#Pos) type, but it's completely manual where `pos` needs to be set for [each type](https://pkg.go.dev/go/ast#StructType). I think the formatters do this manually, as seen [here](https://github.com/golang/go/blob/master/src/cmd/gofmt/internal.go#L94).

main.go
```diff
@@ -20,13 +20,13 @@ import (
...
 	"os"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
```
arjunrn, line 29: This should be immediately after the comment with nothing surrounding it. It's tied to the comment.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -33,12 +35,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
-
-	albo "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
```
arjunrn, line 37: Please maintain the ordering of the imports.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,181 @@
+package awsloadbalancercontroller
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	albo "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
```
arjunrn, line 8: Can you group the imports are described earlier.

thejasn: My bad, forgot about the other files. Fixed across the files.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return true, cr, nil
+}
+
+func (r *AWSLoadBalancerControllerReconciler) ensureCredentialsRequest(ctx context.Context, namespace string, _ *albo.AWSLoadBalancerController) error {
+
+	credReq := createCredentialsRequestName(credentialRequestName)
+
+	reqLogger := log.FromContext(ctx).WithValues("credentialsrequest", credReq)
+	reqLogger.Info("reconciling credentials secret for aws-load-balancer instance")
```
alebedev87, line 43:
```suggestion
	reqLogger.Info("ensuring credentials secret for aws-load-balancer instance")
```

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := r.Client.Get(ctx, name, cr); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+	return true, cr, nil
+}
+
+func (r *AWSLoadBalancerControllerReconciler) ensureCredentialsRequest(ctx context.Context, namespace string, _ *albo.AWSLoadBalancerController) error {
```
alebedev87, line 38: Why a parameter for the controller instance if we don't use it?

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+		_, _, err = r.currentCredentialsRequest(ctx, credReq)
+		return err
+	}
+
+	return nil
+}
+
+func (r *AWSLoadBalancerControllerReconciler) createCredentialsRequest(ctx context.Context, desired *cco.CredentialsRequest) error {
+	if err := r.Client.Create(ctx, desired); err != nil {
+		return fmt.Errorf("failed to create aws-load-balancer credentials request %s: %w", desired.Name, err)
```
alebedev87, line 81: The approach to the error reporting + logging for this method differs from the others (`currentCR`, `updateCR`) where: the errors from the underlying callers are returned and a log message complements the error.
Would be nice to align this to the other methods.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return nil
+}
+
+func (r *AWSLoadBalancerControllerReconciler) updateCredentialsRequest(ctx context.Context, current *cco.CredentialsRequest, desired *cco.CredentialsRequest) (bool, error) {
```
alebedev87, line 86: Would be nice to have a comment to be able to easily find out the meaning of the returned `bool`.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+	return credentialsRequest, nil
+}
+
+func createProviderConfig(codec *cco.ProviderCodec) (*runtime.RawExtension, error) {
+	return codec.EncodeProviderSpec(&cco.AWSProviderSpec{
+		StatementEntries: GetIAMPolicy().Statement,
+	})
+}
+
+// The credentials request will always default to the fixed namespace, so as to
```
alebedev87, line 136: The comment should start from the function name to be a valid `godoc` one.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -38,7 +40,11 @@ import (
 	"github.com/openshift/aws-load-balancer-operator/pkg/aws"
 )
 
-const allowedResourceName = "cluster"
+const (
+	controllerName               = "cluster"
+	controllerSecretName         = "aws-load-balancer-operator-credentials"
+	controllerServiceAccountName = "aws-load-balancer-operator-controller-manager"
```
alebedev87, line 46: These seem to be the operand secret and service account, it's slightly misleading to see `operator` in their names.

pkg/controllers/utils/test/events.go
```diff
@@ -0,0 +1,168 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+
+	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	albv1aplha1 "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
+)
+
+// Event is a simplified representation of the watch event received from the controller runtime client.
```
alebedev87, line 37: This test utility would need to be put somewhere to be added as a dependency as it's now used in more than 1 repository. Something for follow up PRs.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -88,6 +88,18 @@ spec:
...
+          resources:
+          - serviceaccounts
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
```
alebedev87, line 102: BTW why these permissions? I didn't notice the service account manipulation.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
thejasn: fixed

pkg/controllers/awsloadbalancercontroller/credentials_request.go
thejasn: fixed

pkg/controllers/awsloadbalancercontroller/credentials_request.go
thejasn: fixed

pkg/controllers/awsloadbalancercontroller/credentials_request.go
thejasn: Added missing owner reference func call.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
thejasn: fixed

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -0,0 +1,184 @@
...
+// make it future proof. The credentials operator will have limitations in the future, wrt watched namespaces.
+func createCredentialsRequestName(name string) types.NamespacedName {
+	return types.NamespacedName{
+		Name:      name,
+		Namespace: credentialRequestNamespace,
+	}
+}
+
+func createCredentialsSecretRef(operandNamespace string) corev1.ObjectReference {
```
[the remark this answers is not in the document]
thejasn: fixed

pkg/controllers/awsloadbalancercontroller/controller.go
thejasn: Fixed in accordance with [this](https://github.com/openshift/aws-load-balancer-operator/pull/10#discussion_r819339071)

pkg/controllers/utils/test/events.go
thejasn: Noted.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
thejasn: Removed them, it will be added in the next PR.

Makefile
```diff
@@ -62,6 +62,24 @@ CONTAINER_ENGINE ?= docker
...
+IAMCTL_OUTPUT_DIR ?= ./pkg/controllers/awsloadbalancercontroller
+
+# Generated file name.
+IAMCTL_OUTPUT_FILE ?= iam_policy.go
+
+# Go Package of the generated file.
+IAMCTL_GO_PACKAGE ?= awsloadbalancercontroller
+
+# Built go binary path.
+IAMCTL_BINARY ?= ./bin/iamctl
```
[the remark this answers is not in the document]
thejasn: > 
- Why in this repository? it maybe useful for others and isn't tight to aws-lb-operator, so I think it deserves a dedicated repository
- if assets/iam-policy.json turned into pkg/controllers/awsloadbalancercontroller/iam_policy.go with all the asterisks. Was it worth the effort? the latter file could have been just written once without all the generation in the middle


I was initially thinking we could add it [here](https://github.com/openshift/cloud-credential-operator/blob/master/docs/ccoctl.md) when the utility is fleshed out a bit more (not just generate go code, maybe even `CredentialsRequest` too). But we hit [policy size limitations of AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length) which was partly caused by [this](https://issues.redhat.com/browse/CCO-195) discrepancy, so we kind of ended giving up the granularity (which was unfortunate). 

I agree, the utility doesn't make much sense now anymore. One thing we can add as a functionality later on is we sync the policy is `assets` from upstream to ensure it is up-to-date. (similar to what was done for the CRDs) and also can be done rbac as well and thereby utilizing `assets` folder. Anyway, I'm open to suggestions.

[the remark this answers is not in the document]
alebedev87: I think that keeping the code of `iamctl` brings unnecessary heaviness to this repository. The repository gets bigger, the code coverage surface get larger while not bringing a lot of business values. As a compromise I can propose to host it under your personal github and run it with `go run` in here until we find a better place for it. WDYT?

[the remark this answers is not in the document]
thejasn: Yeah, we can do that. @arjunrn what do you think?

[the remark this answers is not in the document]
arjunrn: I would rather have it under this repository for now. Unfortunately we could not use the tool to generate a fine-grained policy. But now that it's in there we can experiment with various minimization techniques to get a better policy. As for coverage, it's a small tool and it won't affect the coverage numbers much. I wanted to ask for tests but I thought we should first get this merged.

PR #14, opened 2022-03-04
Deployment ensure function

arjunrn:
When a deployment is present the controller with ensure that its spec matches the desired state. When it's not present the deployment will be created.


pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -2,13 +2,207 @@ package awsloadbalancercontroller
...
+		err = r.createDeployment(ctx, desired)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create deployment %s: %w", deploymentName, err)
+		}
+		_, current, err = r.currentControllerDeployment(ctx, deploymentName, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get new deployment %s: %w", deploymentName, err)
+		}
+	}
+	return r.updateDeployment(ctx, current, desired)
```
alebedev87, line 48: A call to `currentControllerDeployment` is better to be done if the deployment was updated, similar to what you did for the creation.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -0,0 +1,386 @@
...
+				"--enable-wafv2=false",
+				"--ingress-class=alb",
+			),
+		},
+		{
+			name: "resource tags specified",
+			controller: &albo.AWSLoadBalancerController{
+				Spec: albo.AWSLoadBalancerControllerSpec{
+					AdditionalResourceTags: map[string]string{
+						"test-key": "test-value",
```
alebedev87, line 111: Can you add a couple of tags, just to check the joining of them too?

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -0,0 +1,386 @@
...
+
+			if len(expectedArgs) != len(args) {
+				t.Errorf("expected %d args, instead got %d", len(expectedArgs), len(args))
+				return
+			}
+			expected := expectedArgs.List()
+			sort.Strings(expected)
+			if !cmp.Equal(expected, args) {
+				t.Errorf("unexpected arguments\n%s", cmp.Diff(expected, args))
+			}
```
alebedev87, line 145:
```suggestion
			expected := expectedArgs.List()
			sort.Strings(expected)
			if diff := cmp.Diff(expected, args); diff != "" {
				t.Errorf("unexpected arguments\n%s", diff)
			}
```

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -0,0 +1,386 @@
...
+			).build(),
+			desiredDeployment: testDeployment("operator", "test-namespace", "test-sa").withContainers(
+				testContainer("controller", "controller:v1"),
+			).withReplicas(1).build(),
+			expectedDeployment: testDeployment("operator", "test-namespace", "test-sa").withContainers(
+				testContainer("controller", "controller:v1"),
+			).withReplicas(1).build(),
+		},
+		{
+			name: "container args changed",
```
alebedev87, line 283: Can you please add a test case when an unsolicited container was added or desired removed?

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -0,0 +1,386 @@
...
+			r := &AWSLoadBalancerControllerReconciler{
+				Client: client,
+			}
+			currentDeployment, err := r.updateDeployment(context.Background(), tc.existingDeployment, tc.desiredDeployment)
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+				return
+			}
+
+			if !equality.Semantic.DeepEqual(currentDeployment.Spec, tc.expectedDeployment.Spec) {
```
alebedev87, line 306: What's the advantage of this comparison to `go-cmp`'s one?

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -2,13 +2,207 @@ package awsloadbalancercontroller
...
+	}
+	if !exists {
+		err = r.createDeployment(ctx, desired)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create deployment %s: %w", deploymentName, err)
+		}
+		_, current, err = r.currentControllerDeployment(ctx, deploymentName, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get new deployment %s: %w", deploymentName, err)
+		}
```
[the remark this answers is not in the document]
arjunrn: Don't need it right now because we don't use it to update the status. Will add it in the future.

pkg/aws/aws.go
```diff
@@ -15,31 +17,61 @@ import (
...
 	}
 	region := infra.Status.PlatformStatus.AWS.Region
 
 	awsConfig, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
 	if err != nil {
-		return nil, fmt.Errorf("unable to load AWS config: %w", err)
+		return nil, "", "", fmt.Errorf("unable to load AWS config: %w", err)
+	}
+	ec2Client := ec2.NewFromConfig(awsConfig)
+	infraTagKey := fmt.Sprintf(clusterTagKey, infra.Status.InfrastructureName)
```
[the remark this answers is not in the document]
arjunrn: I checked and it looks like the tag is not added to existing VPCs. But the operator will crash if it doesn't find a suitable VPC. So we will have to ask the users to add the tag.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -2,13 +2,207 @@ package awsloadbalancercontroller
...
+	return r.updateDeployment(ctx, current, desired)
+}
+
+func desiredDeployment(name, namespace, image string, vpcID, clusterName string, controller *albo.AWSLoadBalancerController, sa *corev1.ServiceAccount) *appsv1.Deployment {
+	d := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: appsv1.DeploymentSpec{
```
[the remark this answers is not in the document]
arjunrn: If the users want to modify the placement they can modify the deployment directly. Since that is the approach we are taking for resources I don't see why it shouldn't be the same for placement as well.

pkg/controllers/awsloadbalancercontroller/deployment.go
arjunrn: We don't return the deployment because we are not updating the status with the deployment status. We should do that but not in this PR,

pkg/controllers/awsloadbalancercontroller/deployment_test.go
arjunrn: I'm actually allowing the addition of containers because there could be something else which adds a sidecar and I don't want the operator to stomp that if the desired containers are present.

arjunrn: Will add tests for that however.

pkg/controllers/awsloadbalancercontroller/deployment.go
arjunrn: Wait, I'm wrong, I am returning the deployment. I will update it the code.

pkg/controllers/awsloadbalancercontroller/deployment.go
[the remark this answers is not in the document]
arjunrn: Ignore the previous comment, will update the code.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
arjunrn: Scratch the above as well, I am resetting the containers as well. Looks like I forgot how I implemented everything.

pkg/aws/aws.go
```diff
@@ -15,31 +17,61 @@ import (
...
 	DeleteTags(context.Context, *ec2.DeleteTagsInput, ...func(*ec2.Options)) (*ec2.DeleteTagsOutput, error)
 }
 
-func GetEC2Client(ctx context.Context, client client.Client) (EC2Client, error) {
+type EC2Client interface {
+	VPCClient
+	SubnetClient
+}
+
+func GetEC2Client(ctx context.Context, client client.Client) (EC2Client, string, string, error) {
```
thejasn, line 39: Can you add godoc for the return values?

pkg/controllers/awsloadbalancercontroller/deployment_test.go
arjunrn: My assumption is that equality package is targeted at Kubernetes structs(no idea how) and `cmp.Diff` is more generic. That said I've switched over to `cmp.Diff` in the tests.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -2,13 +2,222 @@ package awsloadbalancercontroller
...
+	}
+	if controller.Spec.Config != nil && controller.Spec.Config.Replicas != 0 {
+		d.Spec.Replicas = pointer.Int32(controller.Spec.Config.Replicas)
+	}
+	return d
+}
+
+func desiredContainerArgs(controller *albo.AWSLoadBalancerController, clusterName, vpcID string) []string {
+	var args []string
+	args = append(args, fmt.Sprintf("--aws-vpc-id=%s", vpcID))
```
thejasn, line 105: I think we need to add [`hostNetwork` arg](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/7baec176ddbaed9d23f92ca9f77a658eed3cf3a3/helm/aws-load-balancer-controller/values.yaml#L201) as well since it will default to `false`

thejasn: Maybe even the `dnsPolicy` field.

thejasn (changes requested):

pkg/controllers/awsloadbalancercontroller/deployment.go
arjunrn: Those are values for the deployment spec

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/7baec176ddbaed9d23f92ca9f77a658eed3cf3a3/helm/aws-load-balancer-controller/templates/deployment.yaml#L46

https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/7baec176ddbaed9d23f92ca9f77a658eed3cf3a3/helm/aws-load-balancer-controller/templates/deployment.yaml#L50

thejasn: Got it, so doesn't it need to be added under our desired deployment?

arjunrn: No it's only for clusters where the pods cannot be given AWS credentials that one would need something like that. When the controller runs on the host network it can retrieve credentials from the [AWS metadata endpoint](https://stackoverflow.com/questions/41024362/aws-retrieving-security-credentials-from-instance-metadata). But we don't do that so it's not required.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -0,0 +1,415 @@
...
+				t.Fatalf("unexpected error: %v", err)
+			}
+			currentDeployment := &appsv1.Deployment{}
+			err = r.Get(ctx, types.NamespacedName{Namespace: tc.expectedDeployment.Namespace, Name: tc.expectedDeployment.Name}, currentDeployment)
+			if err != nil {
+				t.Fatalf("failed to get existing deployment: %v", err)
+			}
+
+			if diff := cmp.Diff(currentDeployment.Spec, tc.expectedDeployment.Spec); diff != "" {
+				t.Fatalf("deployment is not has unexpected fields:\n%s", diff)
```
alebedev87, line 338: `deployment mismatch:\n%s`

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -2,13 +2,222 @@ package awsloadbalancercontroller
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil, nil
+		}
+		return false, nil, err
+	}
+	return true, &deployment, nil
+}
+
+func (r *AWSLoadBalancerControllerReconciler) createDeployment(ctx context.Context, deployment *appsv1.Deployment) error {
```
alebedev87, line 158: Not a big deal, but just noticed a small discrepancy in the naming of the methods: `ensureControllerDeployment` and `currentControllerDeployment` but `createDeployment` and `updateDeployment`.

PR #19, opened 2022-03-09
Add functionality to ensure webhook

arjunrn:


pkg/controllers/awsloadbalancercontroller/webhooks.go
```diff
@@ -2,12 +2,371 @@ package awsloadbalancercontroller
...
+
+func desiredValidatingWebhookConfiguration(controller *albo.AWSLoadBalancerController, webhookService *corev1.Service) *arv1.ValidatingWebhookConfiguration {
+	return &arv1.ValidatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("%s-%s", controllerResourcePrefix, controller.Name),
+		},
+		Webhooks: []arv1.ValidatingWebhook{
+			{
+				Name: "vtargetgroupbinding.elbv2.k8s.aws",
+				ClientConfig: arv1.WebhookClientConfig{
```
alebedev87, line 98: I don't see `caBundle` for the webhooks, the communication is plain http?
As we are not limited by the vanilla Kube support we can use OpenShift's [ca injector](https://docs.openshift.com/container-platform/4.10/security/certificates/service-serving-certificate.html#add-service-certificate-validating-webhook_service-serving-certificate) for the webhookConfigs and [serving cert secret](https://docs.openshift.com/container-platform/4.10/security/certificates/service-serving-certificate.html#add-service-certificate-validating-webhook_service-serving-certificate) for the service.

pkg/controllers/awsloadbalancercontroller/webhooks.go
```diff
@@ -2,12 +2,371 @@ package awsloadbalancercontroller
...
+
+func failurePolicyPtr(failurePolicyType arv1.FailurePolicyType) *arv1.FailurePolicyType {
+	return &failurePolicyType
+}
+
+func (r *AWSLoadBalancerControllerReconciler) updateValidatingWebhookConfiguration(ctx context.Context, current, desired *arv1.ValidatingWebhookConfiguration) error {
+	updatedVWC := current.DeepCopy()
+	var updated bool
+
+	if !haveValidatingWebhooksChanged(updatedVWC.Webhooks, desired.Webhooks) {
```
alebedev87, line 171: Seems like the method name means the opposite of what it's supposed to.

pkg/controllers/awsloadbalancercontroller/webhooks.go
```diff
@@ -2,12 +2,371 @@ package awsloadbalancercontroller
...
+		if err != nil {
+			return fmt.Errorf("failed to create ValidatingWebhookConfiguration %q: %w", desiredVWC.Name, err)
+		}
+	} else {
+		err = r.updateValidatingWebhookConfiguration(ctx, currentVWC, desiredVWC)
+		if err != nil {
+			return fmt.Errorf("failed to updated ValidatingWebhookConfiguration %s: %w", currentVWC.Name, err)
+		}
+	}
+
```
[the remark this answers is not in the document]
arjunrn: Added some logs. At least not the log package included in the controller runtime library.

pkg/controllers/awsloadbalancercontroller/webhooks.go
arjunrn: Updated with the functionality mentioned.

pkg/controllers/awsloadbalancercontroller/webhooks.go
arjunrn: Refactored.

pkg/controllers/awsloadbalancercontroller/service.go
```diff
@@ -84,6 +91,13 @@ func (r *AWSLoadBalancerControllerReconciler) updateService(ctx context.Context,
 		updatedService.Spec.Type = desired.Spec.Type
 	}
 
+	for annotationKey, annotationValue := range desired.Annotations {
+		if currentAnnotationValue, ok := updatedService.Annotations[annotationKey]; !ok || currentAnnotationValue != annotationValue {
+			updatedService.Annotations = desired.Annotations
+			break
+		}
+	}
```
alebedev87, line 99: So, it's possible to have unsolicited annotations but only until any desired one is changed.
Not very strict policy against unsolicited annotations: they are accepted at the edit time but may be gone anytime after.

Maybe a strict policy like `DeepEqual` check of current vs desired instead (like for the containers in the deployment)? Or reset only the desired ones?

arjunrn: That's true, I should not reset the annotations instead add it to the existing list. Faulty logic. Annotations can be added by third party controllers and we don't want to erase them.

pkg/controllers/awsloadbalancercontroller/webhooks.go
```diff
@@ -2,12 +2,371 @@ package awsloadbalancercontroller
...
 	return nil
 }
+
+func (r *AWSLoadBalancerControllerReconciler) currentValidatingWebhookConfiguration(ctx context.Context, name string) (*arv1.ValidatingWebhookConfiguration, bool, error) {
+	var currentVWC arv1.ValidatingWebhookConfiguration
+	err := r.Get(ctx, types.NamespacedName{Name: name}, &currentVWC)
+	if err != nil && errors.IsNotFound(err) {
+		return nil, false, nil
+	}
+	return &currentVWC, true, err
```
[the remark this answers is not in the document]
arjunrn: I think I marked this as Resolved without an explanation(or maybe I updated the code). The err is returned and checked checked in the calling function which should cover that case.

alebedev87 (commented):
LGTM, just the very last comment :)

pkg/controllers/awsloadbalancercontroller/webhooks.go
```diff
@@ -2,12 +2,412 @@ package awsloadbalancercontroller
...
+
+func failurePolicyPtr(failurePolicyType arv1.FailurePolicyType) *arv1.FailurePolicyType {
+	return &failurePolicyType
+}
+
+func (r *AWSLoadBalancerControllerReconciler) updateValidatingWebhookConfiguration(ctx context.Context, current, desired *arv1.ValidatingWebhookConfiguration) error {
+	updatedVWC := current.DeepCopy()
+	var updated bool
+
+	updatedVWC.Annotations, updated = updateAnnotations(updatedVWC.Annotations, desired.Annotations)
```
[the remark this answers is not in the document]
arjunrn: I deliberately avoided doing it that way so that the effects of the function are clear.

PR #23, opened 2022-03-11
Added AWS_DEFAULT_REGION environment variable

arjunrn:
The AWS client requires a default AWS region to be specified. Added functionality to ensure this environment varible is always present in the deployment.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -4,21 +4,15 @@ import (
 	"context"
 	"fmt"
 
-	"k8s.io/apimachinery/pkg/types"
-
-	configv1 "github.com/openshift/api/config/v1"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	clusterInfrastructureName = "cluster"
-	clusterTagKey             = "kubernetes.io/cluster/%s"
-	tagKeyFilterName          = "tag-key"
+	clusterTagKey    = "kubernetes.io/cluster/%s"
+	tagKeyFilterName = "tag-key"
 )
 
 // VPCClient can be used to query VPCs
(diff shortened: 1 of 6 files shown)


pkg/aws/aws.go
```diff
@@ -39,26 +39,37 @@ type EC2Client interface {
 	SubnetClient
 }
 
-// GetEC2Client returns and EC2Client along with the VPC id and the cluster name
-func GetEC2Client(ctx context.Context, client client.Client) (EC2Client, string, string, error) {
+// GetEC2Client returns and EC2Client along with the VPC id, cluster name and AWS region
+func GetEC2Client(ctx context.Context, client client.Client) (ec2Client EC2Client, vpcID string, clusterName string, awsRegion string, err error) {
```
thejasn, line 43: I think we can have additional methods `GetAWSRegion()`,  etc.  defined on the interface and implemented on struct to share required objects, slightly more decoupled. WDYT?

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -224,5 +232,31 @@ func hasContainerChanged(current, desired corev1.Container) (corev1.Container, b
...
+		for _, e := range desired.Env {
+			if ce, ok := currentEnvs[e.Name]; !ok {
+				updated = true
+				break
+			} else {
+				if !cmp.Equal(ce, e) {
+					updated = true
+					break
+				}
+			}
```
thejasn, line 247:
```suggestion
			} else if !cmp.Equal(ce, e) {
				updated = true
				break
			}
```

pkg/aws/aws.go
arjunrn: So the implementing struct is not owned by us. It's actually the EC2 client returned by the SDK. Unfortunately there's no easy way to mock the client. What I have implemented is how they recommend the client be mocked. That said I've refactored the code so that all the information is not returned from a single function. I've split it into 3 functions and added tests for one of them.

thejasn: I actually meant in a new struct, like this 
```go
type AWS struct {
	infra       configv1.Infrastructure
	clustername string
}

func (a AWS) New() AWS {
 // init clustername, etc
}

func (a AWS) GetEC2Client() (EC2Client, error) {
	panic("")
}

func (a AWS) GetVPC() string {
	panic("")
}

func (a AWS) GetRegion() string {
	panic("")
}
```
But splitting it up works as well, I think. For tests I think we can use `gomock` and generate something like [this](https://github.com/openshift/aws-load-balancer-controller/tree/main/mocks/controller-runtime/client) (kind of an overkill for this though).

PR #25, opened 2022-03-15
Mount AWS credentials from CredentialsRequest

arjunrn:
Return the secret which is referenced in the `CredentialsRequest` from the ensure function. Pass this to the ensure function for the deployment so that the secret can be mounted in the pod.

### DIFF
--- a/pkg/controllers/awsloadbalancercontroller/credentials_request.go
+++ b/pkg/controllers/awsloadbalancercontroller/credentials_request.go
@@ -20,7 +20,6 @@ import (
 )
 
 const (
-	credentialRequestName      = "aws-load-balancer-credentials-request"
 	credentialRequestNamespace = "openshift-cloud-credential-operator"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -86,23 +84,23 @@ func (r *AWSLoadBalancerControllerReconciler) createCredentialsRequest(ctx conte
 }
 
 // updateCredentialsRequest updates the credentialrequest if needed and returns a flag to denote if the update was done.
```
thejasn, line 86:
```suggestion
// updateCredentialsRequest updates the credentialrequest if needed.
```

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -36,46 +35,45 @@ func (r *AWSLoadBalancerControllerReconciler) currentCredentialsRequest(ctx cont
 	return true, cr, nil
 }
 
-func (r *AWSLoadBalancerControllerReconciler) ensureCredentialsRequest(ctx context.Context, namespace string, controller *albo.AWSLoadBalancerController) error {
-	credReq := createCredentialsRequestName(credentialRequestName)
+// ensureCredentialsRequest ensures the CredentialsRequest resource and return the secret where the credentials will be written
+func (r *AWSLoadBalancerControllerReconciler) ensureCredentialsRequest(ctx context.Context, namespace string, controller *albo.AWSLoadBalancerController) (string, error) {
+	credReq := createCredentialsRequestName(fmt.Sprintf("%s-%s", controllerResourcePrefix, controller.Name))
```
thejasn, line 40: We have 2 different variables `controllerBaseName` and  `controllerResourcePrefix`, need to reconcile to either one of them.

arjunrn: I've removed the variable `controllerBaseName` for now because it was used in only one location. However there is larger refactor required where we generate the resource names in the individual ensure functions later. But that's outside the scope of this PR.

PR #27, opened 2022-03-16
Refactored resource names

thejasn:

### DIFF
--- a/pkg/controllers/awsloadbalancercontroller/rbac_rolebinding.go
+++ b/pkg/controllers/awsloadbalancercontroller/rbac_rolebinding.go
@@ -20,7 +20,7 @@ import (
 func (r *AWSLoadBalancerControllerReconciler) ensureRoleBinding(ctx context.Context, sa *corev1.ServiceAccount, controller *albo.AWSLoadBalancerController) error {
 	reqLogger := log.FromContext(ctx)
 
-	desired := desiredRoleBinding(ctx, r.Namespace, sa)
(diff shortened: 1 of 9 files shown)


thejasn:
/label tide/merge-method-squash
@arjunrn I'm thinking of adding this to the tide config, WDYT?

pkg/controllers/awsloadbalancercontroller/rbac_rolebinding.go
```diff
@@ -98,8 +98,8 @@ func (r *AWSLoadBalancerControllerReconciler) currentRoleBinding(ctx context.Con
 	return true, obj, nil
 }
 
-func desiredRoleBinding(ctx context.Context, namespace string, sa *corev1.ServiceAccount) *rbacv1.RoleBinding {
-	return buildRoleBinding(commonResourceName, namespace, commonResourceName, sa)
+func desiredRoleBinding(ctx context.Context, name, namespace string, sa *corev1.ServiceAccount) *rbacv1.RoleBinding {
```
arjunrn, line 101: I feel like we should be getting the role name and the service account name from their individual ensure functions and passing them into this function, rather than relying on convention. But we can do that in a followup PR.

arjunrn:
> /label tide/merge-method-squash @arjunrn I'm thinking of adding this to the tide config, WDYT?

Go for it. I think most contributor would use multiple commits in their PRs. So if it's possible to add it into the tide config then we should do it.

PR #32, opened 2022-03-23
Add happy flow testcases

thejasn:


test/e2e/util.go
```diff
@@ -0,0 +1,207 @@
...
+			return false, nil
+		}
+
+		expected := deploymentConditionMap(conditions...)
+		current := deploymentConditionMap(dep.Status.Conditions...)
+		return conditionsMatchExpected(expected, current), nil
+	})
+}
+
+func verifyIngress(t *testing.T, cl client.Client, timeout time.Duration, ingressName types.NamespacedName) (string, error) {
```
arjunrn, line 39: A better name for this would be `getIngressHost()`

test/e2e/operator_test.go
```diff
@@ -158,11 +107,316 @@ func ensureCredentialsRequest() error {
...
+func TestAWSLoadBalancerControllerWithDefaultIngressClass(t *testing.T) {
+	t.Log("Verifying operator availability due to restarts caused by deletion of resources")
+	getOperator(t)
+
+	t.Log("Creating aws load balancer controller instance with default ingress class")
+
+	name := types.NamespacedName{Name: "cluster", Namespace: "aws-load-balancer-operator"}
+	alb := newAWSLoadBalancerController(name, "alb", []albo.AWSAddon{})
+	if err := kubeClient.Create(context.TODO(), &alb); err != nil {
+		if errors.IsAlreadyExists(err) {
```
arjunrn, line 183: We should delete the controller resource after every test. And it should be recreated from scratch in every test. If it already exists then it should not be updated. Because that's a whole different kind of test, where we ensure that updating the configuration yields a properly configured controller deployment.

test/e2e/operator_test.go
```diff
@@ -158,11 +107,316 @@ func ensureCredentialsRequest() error {
...
+		}
+	}
+
 	expected := []appsv1.DeploymentCondition{
 		{Type: appsv1.DeploymentAvailable, Status: corev1.ConditionTrue},
 	}
-	if err := waitForOperatorDeploymentStatusCondition(t, kubeClient, expected...); err != nil {
-		t.Errorf("Did not get expected available condition: %v", err)
+	deploymentName := types.NamespacedName{Name: "aws-load-balancer-controller-cluster", Namespace: "aws-load-balancer-operator"}
+	if err := waitForDeploymentStatusConditionUntilN(t, kubeClient, time.Minute*30, 20, deploymentName, expected...); err != nil {
```
arjunrn, line 333: Does it really have to be 30 minutes? 5 minutes seems reasonable. Setting it to 30 just makes a failing test take longer to fail.

test/e2e/util.go
```diff
@@ -0,0 +1,269 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func waitForDeploymentStatusConditionUntilN(t *testing.T, cl client.Client, timeout time.Duration, count int, deploymentName types.NamespacedName, conditions ...appsv1.DeploymentCondition) error {
+	t.Helper()
+	ctr := 0
+	return wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
```
arjunrn, line 27: Poll every 10 seconds. Every second is really not required.

test/e2e/util.go
```diff
@@ -0,0 +1,269 @@
...
+		dep := &appsv1.Deployment{}
+		if err := cl.Get(context.TODO(), deploymentName, dep); err != nil {
+			t.Logf("failed to get deployment %s: %v (retrying)", deploymentName.Name, err)
+			return false, nil
+		}
+
+		expected := deploymentConditionMap(conditions...)
+		current := deploymentConditionMap(dep.Status.Conditions...)
+		if conditionsMatchExpected(expected, current) {
```
arjunrn, line 36: Are you trying to ensure that the deployment conditions are stable for at least a certain period before returning true?

test/e2e/util.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		return ctr >= count, nil
+	})
+}
+
+func getIngress(t *testing.T, cl client.Client, timeout time.Duration, ingressName types.NamespacedName) (string, error) {
+	t.Helper()
+	var address string
+	return address, wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
```
arjunrn, line 49: Again polling every 1 second is excessive.

test/e2e/util.go
```diff
@@ -0,0 +1,269 @@
...
+	b.ingressclass = class
+	return b
+}
+
+func (b *ingressbuilder) withRules(rules []networkingv1.IngressRule) *ingressbuilder {
+	b.rules = rules
+	return b
+}
+
+func (b ingressbuilder) build() *networkingv1.Ingress {
```
arjunrn, line 230: Using an builder is nice, but the `build()` doesn't have defaults for when the options are not set.

arjunrn (commented):
Overall looks good. I think deleting the CR after every test would really simplify some of the testing logic. You wouldn't have to ensure the deployment is stable for a period if it is always recreated from scratch.

test/e2e/util.go
thejasn: Yes

test/e2e/util.go
thejasn: Fixed.

test/e2e/util.go
thejasn: Fixed.

test/e2e/util.go
thejasn: Fixed.

test/e2e/operator_test.go
```diff
@@ -158,11 +107,316 @@ func ensureCredentialsRequest() error {
...
+
 func TestOperatorAvailable(t *testing.T) {
+	getOperator(t)
+}
+
+// TestAWSLoadBalancerControllerWithDefaultIngressClass tests the basic happy flow for the operator, mostly
+// using the default values.
+func TestAWSLoadBalancerControllerWithDefaultIngressClass(t *testing.T) {
+	t.Log("Verifying operator availability due to restarts caused by deletion of resources")
+	getOperator(t)
```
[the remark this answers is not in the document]
thejasn: Since order can change when someone adds a test before it, figured lets keep it consistent.

test/e2e/operator_test.go
```diff
@@ -158,11 +107,316 @@ func ensureCredentialsRequest() error {
...
+		os.Exit(1)
+	}
+
+	kubeClientSet, err = kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		fmt.Printf("failed to create kube clientset: %s\n", err)
+		os.Exit(1)
+	}
+
+	if err = ensureCredentialsRequest(); err != nil {
```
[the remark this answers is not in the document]
thejasn: Fixed.

test/e2e/util.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		address = ing.Status.LoadBalancer.Ingress[0].Hostname
+		return true, nil
+	})
+}
+
+func waitForDeletion(t *testing.T, cl client.Client, obj client.Object, timeout time.Duration) {
+	t.Helper()
+	deletionPolicy := v1.DeletePropagationForeground
+	_ = wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
```
arjunrn, line 62: I still don't understand why do we have to do this. I think it's not blocking because the child resources don't have the `blockOwnerDeletion` set to true(based on my reading of the [docs](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#DeletionPropagation)).

PR #38, opened 2022-04-05
Parameterize the operand image

arjunrn:
The operand image can be passed as a parameter to the operator. This change adds the image as a environment variable which is copied to the CLI.

### DIFF
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -137,14 +137,6 @@ spec:
         - apiGroups:
           - ""
           resources:
-          - secrets
-          verbs:
-          - get
-          - list
-          - watch
-        - apiGroups:
-          - ""
-          resources:
           - services
           verbs:
           - get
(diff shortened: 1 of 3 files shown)


bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -424,6 +416,8 @@ spec:
                 env:
                 - name: AWS_SHARED_CREDENTIALS_FILE
                   value: /etc/aws-credentials/credentials
+                - name: OPERAND_IMAGE
```
alebedev87, line 419: It makes sense to prefix this variable with `RELATED_IMAGE_` as this will allow us to automatically populate the CSV's `spec.relatedImages` field at CPaaS build time, see [OSBS doc](https://osbs.readthedocs.io/en/latest/users.html#creating-the-relatedimages-section).
This, in turn, will help to enable one of the [main requirements](https://docs.openshift.com/container-platform/4.10/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs) for the disconnected mode.

PR #39, opened 2022-04-07
Docs/e2e tests

thejasn:


arjunrn:
Could you format the file with some line length. 80 or 120, up to you.

arjunrn:
Also include this [file](https://github.com/openshift/release/blob/master/ci-operator/step-registry/aws-load-balancer/install/aws-load-balancer-install-workflow.yaml) in-line in the doc and explain the fields. For some who is completely new to the OpenShift CI that file won't make a lot of sense right away.

PR #43, opened 2022-04-08
Leverage OperatorGroup configuration for permissions

arjunrn:
The namespace which is targetted by the `OperatorGroup` of the operator install configuration is passed to the deployment in the operator CSV through the annotation `olm.targetNamespaces`. The deployment can use this to watch and modify resources in only that namespace.
    
Correspondingly the permissions for the namespaced resources can be restricted to only that namespace.

[OperatorGroup CSV Annotations](https://docs.openshift.com/container-platform/4.10/operators/understanding/olm/olm-understanding-operatorgroups.html#olm-operatorgroups-csv-annotations_olm-understanding-operatorgroups)

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -34,6 +34,7 @@ spec:
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
         - "--image=$(RELATED_IMAGE_CONTROLLER)"
+        - "--namespace=$(TARGET_NAMESPACE)"
         image: controller:latest
         name: manager
         securityContext:
@@ -64,6 +65,10 @@ spec:
             value: /etc/aws-credentials/credentials
           - name: RELATED_IMAGE_CONTROLLER
             value: docker.io/amazon/aws-alb-ingress-controller:v2.4.1
+          - name: TARGET_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.annotations['olm.targetNamespaces']
         volumeMounts:
           - mountPath: /etc/aws-credentials
             name: aws-credentials
--- a/pkg/controllers/awsloadbalancercontroller/controller.go
+++ b/pkg/controllers/awsloadbalancercontroller/controller.go
@@ -67,13 +67,13 @@ type AWSLoadBalancerControllerReconciler struct {
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/finalizers,verbs=update
-//+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=services;secrets,namespace=system,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="networking.k8s.io",resources=ingressclasses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="config.openshift.io",resources=infrastructures,verbs=get;list;watch
-//+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings;clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="apps",resources=deployments,namespace=system,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=serviceaccounts,namespace=system,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,namespace=system,resources=roles;rolebindings,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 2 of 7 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -32,3 +32,5 @@ spec:
         - "--health-probe-bind-address=:8081"
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
+        - "--image=$(RELATED_IMAGE_CONTROLLER)"
+        - "--namespace=$(TARGET_NAMESPACE)"
```
alebedev87, line 36: Why these flags are in the auth proxy patch and not in `config/manager/manager.yaml`?

config/manager/manager.yaml
```diff
@@ -62,6 +62,10 @@ spec:
             value: /etc/aws-credentials/credentials
           - name: RELATED_IMAGE_CONTROLLER
             value: docker.io/amazon/aws-alb-ingress-controller:v2.4.1
+          - name: TARGET_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.annotations['olm.targetNamespaces']
```
alebedev87, line 68: Is there an advantage of using the OLM specific convention over the universal `fieldPath: metadata.namespace`?

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -67,13 +67,14 @@ type AWSLoadBalancerControllerReconciler struct {
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/finalizers,verbs=update
-//+kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="",resources=services,namespace=system,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=secrets,namespace=system,verbs=get;list;watch
```
alebedev87, line 71:
```suggestion
//+kubebuilder:rbac:groups="",resources=services;serviceaccounts,namespace=system,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups="",resources=secrets,namespace=system,verbs=get;list;watch
```
To shorten the resulting output YAML, the line for the service accounts below is not needed then.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -67,13 +67,14 @@ type AWSLoadBalancerControllerReconciler struct {
...
 //+kubebuilder:rbac:groups="networking.k8s.io",resources=ingressclasses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="config.openshift.io",resources=infrastructures,verbs=get;list;watch
-//+kubebuilder:rbac:groups="apps",resources=deployments,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings;clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="apps",resources=deployments,namespace=system,verbs=get;list;watch;create;update;patch;delete
```
alebedev87, line 74: BTW, why `system` namespace? Why not something specific to ALBO?

pkg/controllers/awsloadbalancercontroller/rbac_rules.go
```diff
@@ -35,11 +35,6 @@ func getControllerRules() []rbacv1.PolicyRule {
 			Resources: []string{"pods/status"},
 			Verbs:     []string{"patch", "update"},
 		},
-		{
-			APIGroups: []string{""},
-			Resources: []string{"secrets"},
-			Verbs:     []string{"get", "list", "watch"},
-		},
```
alebedev87, line 42: Why the operand doesn't need this permissions anymore?

config/default/manager_auth_proxy_patch.yaml
arjunrn: Because that get's overriden and I haven't found a way to merge them.

config/manager/manager.yaml
arjunrn: The target namespace may not be the same as the namespace in which the operator is installed.

pkg/controllers/awsloadbalancercontroller/controller.go
arjunrn: No, it's from [kustomize](https://github.com/openshift/aws-load-balancer-operator/blob/c0f9819d0057f5db7c61a7837c41f25b886b77b2/config/manager/manager.yaml#L6). That namespace gets replaced in all the manifests.

pkg/controllers/awsloadbalancercontroller/rbac_rules.go
arjunrn: This is somewhat unrelated. But the upstream dropped those permissions and the operand was still trying to create the cluster role with that permission which caused it to fail.

config/default/manager_auth_proxy_patch.yaml
alebedev87: You can put all the flags of `manager` container in `config/manager/manager.yaml` including the ones from this PR:
```yaml
        args:
        - "--health-probe-bind-address=:8081"
        - "--metrics-bind-address=127.0.0.1:8080"
        - "--leader-elect"
        - "--image=$(RELATED_IMAGE_CONTROLLER)"
        - "--namespace=$(TARGET_NAMESPACE)"
```
Because the ones which are currently in the auth proxy patch should be in the manager regardless the fact if there is a proxy in front or not (leader election, health probe address). Even the metrics address makes sense without the proxy.

arjunrn: Was done in #45

config/manager/manager.yaml
alebedev87: As discussed, most of the time the target namespace will be the same as the operator installation namespace. Unless the `operatorGroup` is manually modified with another namespace whose existence has to be guaranteed by the user.

Though I think I'm fine with the flexibility and with the concept of the rolebinding ensured by OLM in the target namespaces.

PR #44, opened 2022-04-11
Added e2e test for waf-regional and wafv2

thejasn:


test/e2e/operator_test.go
```diff
@@ -220,12 +246,35 @@ func TestAWSLoadBalancerControllerWithDefaultIngressClass(t *testing.T) {
...
+					return nil
+				} else {
+					return err
+				}
+			})
+		if err != nil {
+			t.Fatalf("failed verify ingress with external client: %v", err)
+		}
+
 		defer func() {
```
arjunrn, line 276: I just noticed that this defer is called in a for loop which means all of the deferred functions only get called when the surrounding function exits. But this is unrelated to this change.

test/e2e/operator_test.go
```diff
@@ -338,6 +387,299 @@ func TestAWSLoadBalancerControllerWithCustomIngressClass(t *testing.T) {
...
+			}
+			scanner := bufio.NewScanner(readCloser)
+			defer func() {
+				if err := readCloser.Close(); err != nil {
+					t.Fatalf("failed to close reader for pod %s: %v", clientPod.Name, err)
+				}
+			}()
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.Contains(line, "403 Forbidden") {
```
arjunrn, line 539: I'm not sure that the presence of a 403 Forbidden is a good indicator that things are working as expected. If you really want to verify that the WAF is the reason for this, then the WAF rules should permit requests with only a certain header values and then you need to check with requests with and without those header values. Also the curl pod is just really not required. It's just adding bulk to the tests with no added benefit. WDYT?

test/e2e/operator_test.go
```diff
@@ -338,6 +387,299 @@ func TestAWSLoadBalancerControllerWithCustomIngressClass(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("failed to observe the expected log message: %v", err)
+		}
+		defer func() {
+			waitForDeletion(t, kubeClient, clientPod, defaultTimeout)
+		}()
+	}
+}
+
+func TestAWSLoadBalancerControllerWithWAFv1(t *testing.T) {
```
arjunrn, line 554: So you're not trying to do a direct HTTP request in this test. Why not?

arjunrn (commented):
A couple of questions, but looks good. We should refactor the tests and make it easier to understand for people looking at it for the first time.

test/e2e/operator_test.go
thejasn: > I'm not sure that the presence of a 403 Forbidden is a good indicator that things are working as expected.

Regarding the 403 status check, WAFv2 provides a way to return a custom response when blocked, what would you say about using that? 

> Also the curl pod is just really not required. It's just adding bulk to the tests with no added benefit. WDYT?

The main benefit I see is with the amount of retries and the duration. Currently, it takes about 2-3min for the DNS to get registered and working properly. We can replicate the same with just HTTP clients, just with a few go routines (to handle multiple rules in the future). Also, we are missing an internal load balancer, which I was going to add as well.

thejasn: I'm open to retaining it only for the internal load balancer and replacing the rest. Although, would be better in a different PR.

test/e2e/aws.go
```diff
@@ -0,0 +1,52 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type awsTestHelper struct {
```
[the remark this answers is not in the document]
thejasn: Will simplify it. But looks the credentials don't have enough permission either. 

`    operator_test.go:475: failed to create aws wafv2 acl due to operation error WAFV2: CreateWebACL, https response error StatusCode: 400, RequestID: aca0d697-3801-422d-9d3e-b67d3b6bff52, api error AccessDeniedException: User: arn:aws:iam::460538899914:user/origin-ci-robot-provision is not authorized to perform: wafv2:CreateWebACL on resource: arn:aws:wafv2:us-west-2:460538899914:regional/webacl/echoserver-acl/* because no identity-based policy allows the wafv2:CreateWebACL action`

and 

`     operator_test.go:575: Ensuring test workload namespace aws-load-balancer-test-wafv2
    operator_test.go:598: failed to get change token for waf classic operation error WAF: GetChangeToken, https response error StatusCode: 400, RequestID: 30dc3a79-2347-4e34-9b60-61e4225d2a0c, api error AccessDeniedException: User: arn:aws:iam::460538899914:user/origin-ci-robot-provision is not authorized to perform: waf:GetChangeToken on resource: arn:aws:waf::460538899914:changetoken/* because no identity-based policy allows the waf:GetChangeToken action `

PR #47, opened 2022-04-11
rbac: operand role moved to the bundle

alebedev87:
Reducing the cluster level permissions: 
- cluster role CRUD verbs removed
- cluster role `bind` verb added but only for the operand cluster role
- operand cluster role is now created at installation


config/rbac/kustomization.yaml
```diff
@@ -8,7 +8,6 @@ resources:
 - role.yaml
 - role_binding.yaml
 - upstream_role.yaml
-- upstream_role_binding.yaml
```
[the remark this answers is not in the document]
alebedev87: Done

pkg/controllers/awsloadbalancercontroller/rbac.go
```diff
@@ -9,12 +9,7 @@ import (
 )
 
 func (r *AWSLoadBalancerControllerReconciler) ensureClusterRoleAndBinding(ctx context.Context, sa *corev1.ServiceAccount, controller *albo.AWSLoadBalancerController) error {
-	err := r.ensureClusterRole(ctx, controller)
```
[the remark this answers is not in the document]
alebedev87: Yeah, I think it makes sense. Added the check for the existence of the cluster role.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -73,7 +73,8 @@ type AWSLoadBalancerControllerReconciler struct {
 //+kubebuilder:rbac:groups="apps",resources=deployments,namespace=system,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=serviceaccounts,namespace=system,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=rbac.authorization.k8s.io,namespace=system,resources=roles;rolebindings,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles,verbs=bind,resourceNames=aws-load-balancer-operator-controller-role
```
arjunrn, line 77: Doesn't the permissions here need to be updated? The controller will need permissions to get the clusterrole as well right?

pkg/controllers/awsloadbalancercontroller/rbac_clusterrole.go
```diff
@@ -2,107 +2,29 @@ package awsloadbalancercontroller
...
-		if !errors.IsAlreadyExists(err) {
-			return fmt.Errorf("failed to create clusterrole %s/%s: %w", role.Namespace, role.Name, err)
-		}
-	}
-
-	return nil
-}
-
-func (r *AWSLoadBalancerControllerReconciler) currentClusterRole(ctx context.Context, role *rbacv1.ClusterRole) (bool, *rbacv1.ClusterRole, error) {
+func (r *AWSLoadBalancerControllerReconciler) currentClusterRole(ctx context.Context, name string) (bool, *rbacv1.ClusterRole, error) {
```
arjunrn, line 21: Do we really need 2 different functions: `currentClusterRole` and `verifyClusterRole`. Couldn't the `currentClusterRole` be used directly in the reconcile function?

pkg/controllers/awsloadbalancercontroller/controller.go
alebedev87: Yes they do, forgot to do it. Hm, I'm wondering how e2e didn't catch that.

pkg/controllers/awsloadbalancercontroller/rbac_clusterrole.go
alebedev87: Kept the single method

PR #48, opened 2022-04-12
Vendor controller-gen, kustomize, setup-envtest and yq

arjunrn:
All these tools are now vendor and run from the locally built versions to prevent version differences on different machines.


thejasn:
@arjunrn I think we should vendor `operator-sdk` as well. WDYT?

arjunrn:
> @arjunrn I think we should vendor `operator-sdk` as well. WDYT?

I would have like to done that but the vendoring of operator-sdk is a difficult. It uses some outdated transitive dependencies and I didn't want to pin any versions in our dependencies.

PR #51, opened 2022-04-18
CFE-387: Added support for sts clusters

thejasn:


pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -50,6 +51,8 @@ const (
 	controllerWebhookPort = 9443
 	// common prefix for all resource of an operand
 	controllerResourcePrefix = "aws-load-balancer-controller"
+	// controllerReEnqueueDuration is the delay to re-enqueue.
+	controllerReEnqueueDuration = time.Second * 30
```
arjunrn, line 55: To be precise this value is used when the secret is missing, for some other retry condition it could be something else. So a better name would be `secretMissingReEnqueueDuration`.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -135,6 +138,20 @@ func (r *AWSLoadBalancerControllerReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, fmt.Errorf("failed to ensure CredentialsRequest for AWSLoadBalancerController %q: %w", req.Name, err)
 	}
 
+	secretProvisioned, err := r.ensureCredentialsRequestSecret(ctx, credentialsRequest)
```
arjunrn, line 141: So we are not ensuring the credentials secret here. We are just checking if it is available. So something like `provisioned, err := r.credentialsSecretProvisioned()` would be a better fit. Also if you get a `IsNotFound` error then just return false. It keeps the outer calling function cleaner.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -135,6 +138,20 @@ func (r *AWSLoadBalancerControllerReconciler) Reconcile(ctx context.Context, req
...
+		// updating CR status
+		if err := r.updateControllerStatus(ctx, lbController, nil, credentialsRequest, secretProvisioned); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update status of AWSLoadBalancerController %q: %w", req.Name, err)
+		}
+		logger.Info("(Retrying) failed to ensure secret from credentials request", "secret", credentialsRequest.Spec.SecretRef.Name)
+
+		// retrying after delay to ensure secret provisioning.
+		return ctrl.Result{RequeueAfter: controllerReEnqueueDuration}, nil
+	} else if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to ensure AWSLoadBalancerController %q service account: %w", req.Name, err)
```
arjunrn, line 152: This error message is wrong.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -147,6 +164,21 @@ func desiredDeployment(name, namespace, image, vpcID, clusterName, awsRegion, cr
 								},
 							},
 						},
+						{
+							Name: boundSATokenVolumeName,
+							VolumeSource: corev1.VolumeSource{
+								Projected: &corev1.ProjectedVolumeSource{
+									DefaultMode: aws.Int32(420),
```
arjunrn, line 171: Make this a constant. Also you shouldn't be using the aws pointer helper function. Kubernetes has it's own pointer helper called `pointer.Int32()`.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -147,6 +164,21 @@ func desiredDeployment(name, namespace, image, vpcID, clusterName, awsRegion, cr
...
+									Sources: []corev1.VolumeProjection{{
+										ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+											Audience:          "openshift",
+											ExpirationSeconds: aws.Int64(3600),
```
arjunrn, line 175: Same goes for this.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -562,10 +563,21 @@ func TestEnsureDeployment(t *testing.T) {
...
 				).build(),
 			).withControllerReference("cluster").withVolumes(
 				corev1.Volume{Name: "aws-credentials", VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{SecretName: "test-credentials"}}},
 				corev1.Volume{Name: "tls", VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{SecretName: "test-serving"}}},
+				corev1.Volume{Name: "bound-sa-token", VolumeSource: corev1.VolumeSource{Projected: &corev1.ProjectedVolumeSource{
+					DefaultMode: aws.Int32(420),
+					Sources: []corev1.VolumeProjection{{
+						ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+							Audience:          "openshift",
+							ExpirationSeconds: aws.Int64(3600),
```
arjunrn, line 576: Change this please. Also do we have tests covering any change we make to the bound service account token?

pkg/controllers/awsloadbalancercontroller/serviceaccount.go
```diff
@@ -63,6 +64,7 @@ func desiredAWSLoadBalancerServiceAccount(namespace string, controller *albo.AWS
 			Namespace: namespace,
 			Name:      fmt.Sprintf("%s-%s", controllerResourcePrefix, controller.Name),
 		},
+		AutomountServiceAccountToken: aws.Bool(true),
```
arjunrn, line 67: Now that we are using a non default value we have to update the functions which verifies if the service account needs an update.

pkg/controllers/awsloadbalancercontroller/status.go
```diff
@@ -16,16 +16,21 @@ import (
 )
 
 const (
-	DeploymentAvailableCondition         = "DeploymentAvailable"
-	DeploymentUpgradingCondition         = "DeploymentUpgrading"
-	CredentialsRequestUpdatingCondition  = "CredentialsRequestUpgrading"
-	CredentialsRequestAvailableCondition = "CredentialsRequestAvailable"
+	DeploymentAvailableCondition = "DeploymentAvailable"
+	DeploymentUpgradingCondition = "DeploymentUpgrading"
+	CredentialsRequestAvailable  = "CredentialsRequestAvailable"
```
arjunrn, line 21: This constant should still be called `CredentialsRequestAvailableCondition`.

pkg/controllers/awsloadbalancercontroller/status.go
```diff
@@ -34,42 +39,25 @@ func (r *AWSLoadBalancerControllerReconciler) updateControllerStatus(ctx context
 	return nil
 }
 
-func credentialRequestsConditions(cr *cco.CredentialsRequest, generation int64) []metav1.Condition {
+func credentialRequestsConditions(cr *cco.CredentialsRequest, secretProvisioned bool, generation int64) []metav1.Condition {
 	var conditions []metav1.Condition
```
arjunrn, line 43: I don't see the credentials request being used anywhere. Do we still have to pass it to this function.

arjunrn: Also the condition name now doesn't make sense. We should call it something like `CredentialsSecretAvailable`.

arjunrn (changes requested):
Still needs some work.

pkg/controllers/awsloadbalancercontroller/controller.go
thejasn: > Also if you get a IsNotFound error then just return false. It keeps the outer calling function cleaner.

It doesn't help much, since we still need to handle other types of errors (unless we suppress those errors) 

```go
if err == nil && !secretProvisioned {
       // update status and re-enqueue
} else if err != nil {
      // return non-NotFound errors
}
```

pkg/controllers/awsloadbalancercontroller/controller.go
thejasn: Fixed.

pkg/controllers/awsloadbalancercontroller/controller.go
thejasn: Fixed.

pkg/controllers/awsloadbalancercontroller/deployment.go
thejasn: Fixed.

pkg/controllers/awsloadbalancercontroller/deployment.go
thejasn: Fixed.

pkg/controllers/awsloadbalancercontroller/serviceaccount.go
thejasn: Yeah, my bad, I forgot this. Added.

pkg/controllers/awsloadbalancercontroller/status.go
thejasn: Fixed

pkg/controllers/awsloadbalancercontroller/status.go
thejasn: > I don't see the credentials request being used anywhere. Do we still have to pass it to this function.

Changed argument to be just the secret name. 

> Also the condition name now doesn't make sense. We should call it something like CredentialsSecretAvailable.

Fixed

pkg/controllers/awsloadbalancercontroller/controller.go
arjunrn: All other errors we simply return an error even if `secretProvisioned` is true(because we never expect `secretProvisioned` to be true and for there to be an error). 

```
provisioned, err := r.credentialsSecretProvisioned()
if err!=nil {
   // this is only possible when there is some error other than not found
  // and we want to immediately retry because it's probably transient
  return err
}

r.updateStatus(..provisioned)
```

docs/sts/sts.md
```diff
@@ -0,0 +1,73 @@
+# Using manual mode with STS
+
+In a STS Cluster, *CredentialsRequests* are not automatically provisioned by
+the `cloud-credential-operator` and requires manual intervention by the
+cluster-admin. They need to be provisioned using the `ccoctl` binary.
```
arjunrn, line 5:
```suggestion
cluster-admin. The credentials secret needs to be provisioned using the `ccoctl` binary.
```

docs/sts/sts.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `aws-load-balancer-operator` also relies on the `cloud-credential-operator`
+to provision the *CredentialsRequest* so in an STS Cluster, it needs to be
+provisioned manually. The operator will wait until the required secrets are
+created and available.
```
arjunrn, line 10:
```suggestion
The `aws-load-balancer-operator` also relies on the `cloud-credential-operator`
to provision the secret for *CredentialsRequest*. And so in an STS Cluster the 
secret needs to be provisioned manually. The operator will wait until the required
secrets are created and available.
```

docs/sts/sts.md
```diff
@@ -0,0 +1,73 @@
...
+    ```bash
+    oc image extract $CCO_IMAGE --file="/usr/bin/ccoctl" -a ~/.pull-secret
+    ```
+
+4. Change the permissions to make ccoctl executable:
+
+    ```bash
+    chmod 775 ccoctl
+    ```
+
```
arjunrn, line 40: Let's just link to the existing documentation where they describe how to get _ccoctl_.

docs/sts/sts.md
```diff
@@ -0,0 +1,73 @@
...
+    For each *CredentialsRequest* object, `ccoctl` creates an IAM role with a trust
+    policy that is tied to the specified OIDC identity provider, and a permissions
+    policy as defined in each *CredentialsRequest* object. This also generates a set
+    of manifests which includes the secrets that is required
+    by the **aws-load-balancer-operator**.
+
+3. Apply the secrets to your cluster:
+
+    ```bash
+    ls <path_to_ccoctl_output_dir>/manifests/*-credentials.yaml | xargs -I{} oc apply -f {}
```
arjunrn, line 72: `<path_to_ccoctl_output_dir>` is not defined anywhere.

pkg/controllers/awsloadbalancercontroller/credentials_request.go
```diff
@@ -78,6 +78,19 @@ func (r *AWSLoadBalancerControllerReconciler) ensureCredentialsRequest(ctx conte
 	return current, nil
 }
 
+func (r *AWSLoadBalancerControllerReconciler) credentialsSecretProvisioned(ctx context.Context, cr *cco.CredentialsRequest) (bool, error) {
```
arjunrn, line 81: This function should return a `false, nil` when the error is `IsNotFound` so that the caller doesn't have to handle it.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -24,6 +24,9 @@ const (
 	appLabelName    = "app.kubernetes.io/name"
 	appName         = "aws-load-balancer-operator"
 	appInstanceName = "app.kubernetes.io/instance"
+	// awsSDKLoadConfigName is the name of the environment variable which enables shared configs.
+	// Without which certain fields in the config aren't set. Eg: role_arn.
+	awsSDKLoadConfigName = "AWS_SDK_LOAD_CONFIG"
```
arjunrn, line 29: Did you verify that this is actually required?

thejasn: Yes, https://github.com/aws/aws-sdk-go/blob/v1.43.42/aws/session/shared_config.go#L332 doesn't bind `role_arn` without setting `exOpts` and `AWS_SDK_LOAD_CONFIG` sets that value.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -135,6 +138,23 @@ func (r *AWSLoadBalancerControllerReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, fmt.Errorf("failed to ensure CredentialsRequest for AWSLoadBalancerController %q: %w", req.Name, err)
 	}
 
+	secretProvisioned, err := r.credentialsSecretProvisioned(ctx, credentialsRequest)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to ensure credentialsrequest secret %q for AWSLoadBalancerController %q: %w", credentialsRequest.Spec.SecretRef.Name, req.Name, err)
```
arjunrn, line 143: The error message should be different. Probably _failed to verify if credentials secret %q for AWSLoadBalancerController %q has been provisioned_.

arjunrn, line 143: The error message should be different. Probably _failed to verify if credentials secret %q for AWSLoadBalancerController %q has been provisioned_.

thejasn: Fixed.

PR #57, opened 2022-05-05
Replace kubectl with oc

arjunrn:


thejasn:
Are we replacing everywhere, or just the ones that reflect in the docs?

PR #64, opened 2022-05-16
Bug 2086450: Explicity set the security context for the controller

arjunrn:
The controller is currently running with the default container security context. This change set the field `runAsNonRoot` to _true_ and the fields `privileged` and `allowPrivilegeEscalation` to _false_. It also sets the secomp profile to the default and drops all capabilities.

### DIFF
--- a/pkg/controllers/awsloadbalancercontroller/deployment.go
+++ b/pkg/controllers/awsloadbalancercontroller/deployment.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/utils/pointer"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -45,6 +46,8 @@ const (
 	boundSATokenVolumeName = "bound-sa-token"
 	// boundSATokenDir is the sa token directory
 	boundSATokenDir = "/var/run/secrets/openshift/serviceaccount"
+	// all capabilities in the pod security context
+	allCapabilities = "ALL"
 )
 
 func (r *AWSLoadBalancerControllerReconciler) ensureDeployment(ctx context.Context, namespace, image string, sa *corev1.ServiceAccount, crSecretName, servingSecretName string, controller *albo.AWSLoadBalancerController) (*appsv1.Deployment, error) {
@@ -141,6 +144,17 @@ func desiredDeployment(name, namespace, image, vpcID, clusterName, awsRegion, cr
 									Name:      boundSATokenVolumeName,
 									MountPath: boundSATokenDir,
 									ReadOnly:  true,
+								},
+							},
+							SecurityContext: &corev1.SecurityContext{
+								Capabilities: &corev1.Capabilities{
+									Drop: []corev1.Capability{allCapabilities},
+								},
+								Privileged:               pointer.Bool(false),
+								RunAsNonRoot:             pointer.Bool(true),
+								AllowPrivilegeEscalation: pointer.Bool(false),
+								SeccompProfile: &corev1.SeccompProfile{
+									Type: corev1.SeccompProfileTypeRuntimeDefault,
 								},
 							},
 						},
@@ -285,8 +299,8 @@ func (r *AWSLoadBalancerControllerReconciler) updateDeployment(ctx context.Conte
 				return false, fmt.Errorf("deployment %s does not have a container with the name %s", current.Name, desiredContainer.Name)
 			}
 
-			if container, changed := hasContainerChanged(updated.Spec.Template.Spec.Containers[foundIndex], desiredContainer); changed {
-				updated.Spec.Template.Spec.Containers[foundIndex] = container
+			if changed := hasContainerChanged(updated.Spec.Template.Spec.Containers[foundIndex], desiredContainer); changed {
+				updated.Spec.Template.Spec.Containers[foundIndex] = desiredContainer
 				outdated = true
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -377,3 +395,63 @@ func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
...
+			return true
+		}
+		if desired.SeccompProfile.Type != "" && desired.SeccompProfile.Type != current.SeccompProfile.Type {
+			return true
+		}
+	}
+	return false
+}
+
+func compareBoolPtr(current, desired *bool) bool {
```
alebedev87, line 444:
```suggestion
func equalBoolPtr(current, desired *bool) bool {
```

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -339,6 +353,10 @@ func hasContainerChanged(current, desired corev1.Container) (corev1.Container, b
 		updated = true
 	}
 
+	if hasSecurityContextChanged(current.SecurityContext, desired.SecurityContext) {
+		updated = true
```
alebedev87, line 357: Why not returning right here? The whole container will be reset regardless of the results of the further comparisons.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -647,3 +711,158 @@ func TestEnsureDeployment(t *testing.T) {
...
+			name:      "SeccompProfile is empty",
+			currentSC: &corev1.SecurityContext{SeccompProfile: &corev1.SeccompProfile{}},
+			desiredSC: &corev1.SecurityContext{SeccompProfile: &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault}},
+			changed:   true,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			changed := hasSecurityContextChanged(tc.currentSC, tc.desiredSC)
+			if changed != tc.changed {
+				t.Errorf("expected security has changed to be be %v, instead was %v", tc.changed, changed)
```
alebedev87, line 864:
```suggestion
				t.Errorf("expected security has changed to be %v, instead was %v", tc.changed, changed)
```

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -377,3 +395,63 @@ func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
...
+		}
+		cmpCapabilities := cmpopts.SortSlices(func(a, b corev1.Capability) bool { return a < b })
+		if !cmp.Equal(desired.Capabilities.Add, current.Capabilities.Add, cmpCapabilities) {
+			return true
+		}
+
+		if !cmp.Equal(desired.Capabilities.Drop, current.Capabilities.Drop, cmpCapabilities) {
+			return true
+		}
+	}
```
alebedev87, line 420: what if desired capabilities are `nil` but current are not?

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -377,3 +395,63 @@ func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
...
+
+	if desired.SeccompProfile != nil {
+		if current.SeccompProfile == nil {
+			return true
+		}
+		if desired.SeccompProfile.Type != "" && desired.SeccompProfile.Type != current.SeccompProfile.Type {
+			return true
+		}
+	}
```
alebedev87, line 440: same question as for the capabilities

arjunrn: For both cases we don't want to set the value to `nil`. Ideally we will never have the desired as `nil`. If however in the first pass it gets created with `nil` it may be defaulted to something else due to webhooks. Subsequently, if we set it back to `nil` our controlller will end up constantly changing the value which is defaulted elsewhere.

pkg/controllers/awsloadbalancercontroller/deployment.go
arjunrn: I've add some logic to handle the case for when the desired is `nil` but generally that won't be the case. I would not want the controller to set something to `nil` when it might be being set to some value but a webhook.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -377,3 +390,63 @@ func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
...
+	return false
+}
+
+func equalBoolPtr(current, desired *bool) bool {
+	if desired == nil && current == nil {
+		return false
```
alebedev87, line 441: The `return`s in this function need to be inverted to match the function name.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: Yeah, I think it makes sense.
Then to be consistent, shouldn't we follow the same logic for the bool pointers (`RunAsNonRoot` etc.)? They are currently don't allow defaulting:
```go
		{
			name:      "desired RunAsNonRoot is nil",
			currentSC: &corev1.SecurityContext{RunAsNonRoot: pointer.BoolPtr(false)},
			desiredSC: &corev1.SecurityContext{},
			changed:   true,
		},
```

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -647,3 +711,202 @@ func TestEnsureDeployment(t *testing.T) {
...
+		},
+		{
+			name:      "current Privileged is nil",
+			currentSC: &corev1.SecurityContext{},
+			desiredSC: &corev1.SecurityContext{Privileged: pointer.BoolPtr(false)},
+			changed:   true,
+		},
+		{
+			name:      "desired Privileged is nil",
+			desiredSC: &corev1.SecurityContext{},
```
alebedev87, line 760: This test case results in `changed == true` because `RunAsNonRoot` condition precedes the `Priviledged` condition.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -377,3 +390,63 @@ func indexedContainerEnv(envs []corev1.EnvVar) map[string]corev1.EnvVar {
...
+
+	if current == nil {
+		return false
+	}
+
+	if *current != *desired {
+		return false
+	}
+	return true
+}
```
thejasn, line 452: I think `cmp.Equal()` has the same behavior for pointers as well.

arjunrn: It's not the same behavior. Because this function is returns true when desired is nil but not when current is nil when desired is not. `cmp.Equal()` won't give the same behavior. The reason for the special type of comparison is discussed [here](https://github.com/openshift/aws-load-balancer-operator/pull/64#discussion_r874169605).

alebedev87:
/lgtm

bot, where are you?

thejasn:
@arjunrn Should this change be made to the operator's deployment as well? As in, directly to the CSV.

arjunrn:
@thejasn It's already present in the deployment, over [here](https://github.com/openshift/aws-load-balancer-operator/blob/12bcb4a1b4c473453aef78a82621a1eb407b6c13/config/manager/manager.yaml#L41) and [here](https://github.com/openshift/aws-load-balancer-operator/blob/12bcb4a1b4c473453aef78a82621a1eb407b6c13/config/manager/manager.yaml#L28).

PR #70, opened 2022-06-07
Explicity set the scopes for the webhook rules

arjunrn:
The webhook rules also have a scope field which is defaulted to _all_("*") and this was being flagged as an update required by the
controller. So the scope is now set and the comparision of the fields on the rule is performed on each field individually. Also added tests to verify this functionality.


thejasn:
LGTM

PR #74, opened 2022-10-20
Prepapre for golang 1.19 CI config change

alebedev87:
[Migration to golang 1.19](https://github.com/openshift/release/pull/33283) raised a couple of problems:
- `golangci-lint` of version `v1.44.2` which is currently used in `main` breaks any linting resulting in OOM due to the lack of 1.19 support introduced later (`>=v1.48.0`)
- security context which would comply with `restricted` PodSecurity (enforced in 4.12) is missing in the e2e test POD

`build_root` is added to the repository as well, to be more flexible in the future.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -94,6 +94,12 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
 
 // buildEchoPod returns a pod definition for an socat-based echo server.
 func buildEchoPod(name, namespace string) *corev1.Pod {
+	var (
+		allCapabilities          = "ALL"
+		privileged               = false
+		runAsNonRoot             = true
+		allowPrivilegeEscalation = false
+	)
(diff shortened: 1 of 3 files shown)


test/e2e/util.go
```diff
@@ -94,6 +94,12 @@ func conditionsMatchExpected(expected, actual map[string]string) bool {
 
 // buildEchoPod returns a pod definition for an socat-based echo server.
 func buildEchoPod(name, namespace string) *corev1.Pod {
+	var (
```
Miciah, line 97:
```suggestion
	const (
```

Miciah: Oh, I guess they can't be consts since you need to take their reference.

alebedev87:
/label docs-approved
/label px-apporved
/label qe-approved

Change for CI config golang 1.19 to pass.

Issue #77, opened 2022-11-01
The operator failed to start


datianshi:
Seems the operator does not have a projected token configured.
I wonder how does this work with STS

{"error": "failed to list VPC with tag \"kubernetes.io/cluster/${cluster_name}\": operation error EC2: DescribeVpcs, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, failed to retrieve jwt from provide source, unable to read file at /var/run/secrets/openshift/serviceaccount/token: open /var/run/secrets/openshift/serviceaccount/token: no such file or directory"}

thejasn:
Hey @datianshi,
The operator is supported on STS clusters as well. Both on non-STS cluster and STS cluster have a prerequisite during installation. You can find the documentation for the same [here](https://github.com/openshift/aws-load-balancer-operator/blob/main/docs/prerequisites.md).

datianshi:
@thejasn Understand the STS. I think the issue still exists as there is no projected token configured in the manager. 
I created a PR here -> https://github.com/openshift/aws-load-balancer-operator/pull/78

alebedev87:
Addressed in https://github.com/openshift/aws-load-balancer-operator/pull/82

PR #78, opened 2022-11-02
operator needs projected token for STS credentials

datianshi:
This is related to #77 

{"error": "failed to list VPC with tag "kubernetes.io/cluster/${cluster_name}": operation error EC2: DescribeVpcs, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, failed to retrieve jwt from provide source, unable to read file at /var/run/secrets/openshift/serviceaccount/token: open /var/run/secrets/openshift/serviceaccount/token: no such file or directory"}

The operator( Not the controller) in STS mode requires the projected token. However, the token is not configured

### DIFF
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -333,6 +333,9 @@ spec:
                 volumeMounts:
                 - mountPath: /etc/aws-credentials
                   name: aws-credentials
+                - mountPath: /var/run/secrets/openshift/serviceaccount
+                  name: bound-sa-token
+                  readOnly: true
               securityContext:
                 runAsNonRoot: true
                 seccompProfile:
@@ -346,6 +349,14 @@ spec:
                   - key: credentials
                     path: credentials
                   secretName: aws-load-balancer-operator
+              - name: bound-sa-token
+                projected:
+                  defaultMode: 420
+                  sources:
+                  - serviceAccountToken:
+                      audience: openshift
+                      expirationSeconds: 3600
+                      path: token
       permissions:
       - rules:
         - apiGroups:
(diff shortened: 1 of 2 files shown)


jharrington22:
@datianshi thanks for this PR, we should also need a change to the AWS client to [enable shared](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/) config to load support for assume role. Were you able to test this with just the changes you have here?

jharrington22:
@datianshi I see ALB controller in this operator is adding a projected SA token volume to the deployment [here](https://github.com/openshift/aws-load-balancer-operator/blob/24aa48a8c516e15b6ce9ecea7641e854969f8c54/pkg/controllers/awsloadbalancercontroller/deployment.go#L181-L193). I will try to test the operator as is today on a running ROSA STS cluster.

alebedev87:
@datianshi : `make bundle` would need to be run to replicate the token volume to OLM's bundle.

datianshi:
@alebedev87  any idea why the test failed. I can run make bundle locally without problem

datianshi:
What else we need to merge this PR

alebedev87:
@datianshi : I wanted to ask you to provide examples for IAM resources and the secret which you used but then realized that this maybe a bit too much as some docs would need to be updated too. I took over from your PR to add the examples and the doc as well: https://github.com/openshift/aws-load-balancer-operator/pull/82. Hope this is fine with you.

jharrington22:
@alebedev87 Thanks for updating the documentation for the second CredentialsRequest created by the operator for the controller. Were you able to test the creation of an ALB?

alebedev87:
@jharrington22 : yes but on a simple STS cluster (not ROSA). Results can be found in the description of https://github.com/openshift/aws-load-balancer-operator/pull/82.

alebedev87:
Closing in favor of https://github.com/openshift/aws-load-balancer-operator/pull/82.

PR #79, opened 2022-11-04
CFE-648: Upgrade aws load balancer controller to v2.4.4

thejasn:
Description
---

Bumps aws-load-balancer-controller to `v2.4.4` and the corresponding default controller image  to `quay.io/aws-load-balancer-operator/aws-load-balancer-controller:95d71cc4e7e5` (pins to https://github.com/openshift/aws-load-balancer-controller/tree/2c04703264566be3608cbbf4145929b549fe320a).

Docs and Release notes: https://github.com/openshift/openshift-docs/pull/52292, https://github.com/openshift/openshift-docs/pull/52309

### DIFF
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -299,7 +299,7 @@ spec:
                 - name: AWS_SHARED_CREDENTIALS_FILE
                   value: /etc/aws-credentials/credentials
                 - name: RELATED_IMAGE_CONTROLLER
-                  value: docker.io/amazon/aws-alb-ingress-controller:v2.4.4
+                  value: quay.io/aws-load-balancer-operator/aws-load-balancer-controller@sha256:95d71cc4e7e594fd7cca52c5904c3cf86837bf1e90955eecee6b5cd0862eb501
                 - name: TARGET_NAMESPACE
                   valueFrom:
                     fieldRef:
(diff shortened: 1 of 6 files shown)


thejasn:
Operand image bump, wait until https://github.com/openshift/aws-load-balancer-controller/pull/11

thejasn:
/test e2e-aws-operator
```
operator_test.go:425: failed to create aws wafv2 acl due to operation error WAFV2: CreateWebACL, https response error StatusCode: 400, RequestID: 3c11b418-0e59-46d4-8595-f5b55dad04a7, WAFDuplicateItemException: AWS WAF couldn’t perform the operation because some resource in your request is a duplicate of an existing one.
```
Maybe failure in clean up?

alebedev87:
@thejasn : I think checking for existence won't harm here. If we end up in the situation when Web ACL is already there - we need to remove it anyway, so failing at creation needs to be skipped. PTAL: https://github.com/openshift/aws-load-balancer-operator/pull/81.

thejasn:
rebase to include #81

alebedev87:
/hold
To let NetEdge folks to have a look.

Miciah:
The default value for the `--image` flag still has an old pullspec:
https://github.com/openshift/aws-load-balancer-operator/blob/50017e40bed7766ec1cdc52f2ad880bff9bd2354/main.go#L88
It isn't a critical issue, since the bundle manifest specifies `--image` with an explicit pullspec, so the default value for `--image` only matters if someone manually deploys the operator outside of a supported context.  However it does look weird that the operator still has a reference to the old v2.4.1 tag.  More importantly, once #80 merges, I believe the operator will fail to start aws-load-balancer-controller if the operator is started with the default `--image` and tries to set the `--feature-gates=EnableIPTargetType=false` flag that only exists in newer aws-load-balancer-controller versions, right?  So it would be good to update the default value for `--image` in this PR, or in a follow-up.  

Other than that,
/lgtm

alebedev87:
/unhold
As Miciah gave his LGTM.

alebedev87:
/label px-approved

As per Chris Fields: docs are enough for this ALBO release.

PR #80, opened 2022-11-04
CFE-649: Disable ip target type on Service and Ingress resources

thejasn:
Description
---
In the previous release, the user was allowed to set the `target-type=ip` on `Service` and `Ingress` type resources, which is not supported on Openshift due to its dependency on AWS VPC CNI. A feature gate `EnableIPTargetType` was added to the controller to controller if `target-type=ip` is enabled. 

This PR disables IP mode for `target-type` on both `Service` and `Ingress` resources. 

Release notes: https://github.com/openshift/openshift-docs/pull/52292


thejasn:
Wait for default operand image bump in #79

thejasn:
rebased from `main`

alebedev87:
/label px-approved

As per Chris Fields statement: docs are enough for this release of ALBO.

PR #83, opened 2022-11-23
CFE-619: release 0.2 - add stable-v0.2 and stable-v0 channels

alebedev87:
This PR adds the new channels:

- `stable-v0.2` for the upcoming release
- `stable-v0` for the users ready to streamline the minor version upgrades

`stable-v0` channel would allow the engineering to reduce the configuration burden too as the channel needs to be mentioned in all the docs and the Prow CI configuration.
`stable-v0.1` is kept in `main` for the moment to not break CI as it uses `stable-v0.1` channel to install the operator. A follow-up PR will need to be done to remove it.


alebedev87:
Docs PR: https://github.com/openshift/openshift-docs/pull/53157

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -100,4 +101,4 @@ spec:
   provider:
     name: Red Hat
     url: https://redhat.com
-  version: 0.0.0
+  version: 0.2.0
```
Miciah, line 104: It seems odd that the version was 0.0.0 here.  Was the repository initialized with the correct version in `bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml` but an incorrect version here in `config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml`?

Are we skipping 0.1.0 because version 0.0.1 was shipped in channel stable-v0.1 (instead of in stable-v0.0)?

Miciah (commented):
/lgtm
assuming that skipping the 0.1.0 version is intentional.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: - The discrepancy of the version from `bundle/` and `config/` is the same as [you spotted in external-dns PR](https://github.com/openshift/external-dns-operator/pull/183#discussion_r1030903384). It comes from the generation of the bundle manifests, add the version mentioned Makefile to `bundle/` [during the generation](https://github.com/openshift/aws-load-balancer-operator/blob/main/Makefile#L208). That's why we forget to update the version from `config/` because it's not really used during the bundle generation.
- The version we have in GitHub is used only by the people who build the operator from GitHub. The version shipped to the Red Hat customer gets rendered on CPaaS ([link](https://gitlab.cee.redhat.com/cpaas-midstream/aws-load-balancer-operator/-/blob/aws-lb-optr-0.1-rhel-8/distgit/containers/aws-load-balancer-operator-bundle/render_templates#L107-111)). So far we keep both in sync (for ExtDNS Operator too btw) but technically they can be different.
- There was a mistake made in the versioning of ALBO, the channel was called `stable-v0.1` but the version shipped (on CPaaS and GitHub) was `0.0.1`. We fixed the versioning for `0.2` release: with this PR on GitHub, on CPaaS it's already done.

alebedev87:
/retest

`Failed to pull ci-operator image` O_O.

PR #90, opened 2023-01-23
OCPBUGS-5295: egress proxy support for controller

alebedev87:
- Propagate the proxy environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`) down to the controller.
- Add the custom trusted CA bundle to the controller: new volume and volume mount in the desired controller's deployment.
- Redeploy the controller if the custom CA bundle configmap changes: new watch for the CA bundle configmap in the operator, new annotation in the desired controller's deployment.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -37,6 +37,7 @@ spec:
         - "--leader-elect"
         - "--image=$(RELATED_IMAGE_CONTROLLER)"
         - "--namespace=$(TARGET_NAMESPACE)"
+        - "--trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)"
         image: controller:latest
         name: manager
         securityContext:
@@ -74,6 +75,7 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: TRUSTED_CA_CONFIGMAP_NAME
         volumeMounts:
           - mountPath: /etc/aws-credentials
             name: aws-credentials
--- a/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -6,6 +6,7 @@ metadata:
     capabilities: Basic Install
     olm.skipRange: <0.2.0
     operatorframework.io/suggested-namespace: aws-load-balancer-operator
+    operators.openshift.io/infrastructure-features: '["proxy-aware"]'
     repository: https://github.com/openshift/aws-load-balancer-operator
     support: Red Hat, Inc.
   name: aws-load-balancer-operator.v0.2.0
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,6 +1,8 @@
 # Configuring egress proxy for AWS Load Balancer Operator
 
-If a cluster wide egress proxy is configured on the OpenShift cluster, OLM automatically updates all the operators' deployments with `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` environment variables.
+
+If a cluster wide egress proxy is configured on the OpenShift cluster, OLM automatically updates all the operators' deployments with `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` environment variables.  
+Those variables are then propagated down to the managed controller by the AWS Load Balancer Operator.
 
 ## Trusted Certificate Authority
 
@@ -8,7 +10,7 @@ If a cluster wide egress proxy is configured on the OpenShift cluster, OLM autom
 Follow the instructions below to let AWS Load Balancer Operator trust a custom Certificate Authority (CA). The operator's OLM subscription has to have been created first.
 The operator's deployment doesn't have to be ready though.
 
-1. Create the configmap containing the CA bundle in `aws-load-balancer-operator` namespace.
+1. Create the configmap containing the CA bundle in `aws-load-balancer-operator` namespace. Run the following commands to [inject](https://docs.openshift.com/container-platform/4.12/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki) the CA bundle trusted by OpenShift into a configmap:
     ```bash
     oc -n aws-load-balancer-operator create configmap trusted-ca
     oc -n aws-load-balancer-operator label cm trusted-ca config.openshift.io/inject-trusted-cabundle=true
@@ -17,15 +19,16 @@ The operator's deployment doesn't have to be ready though.
 2. Consume the created configmap in AWS Load Balancer Operator's deployment by updating its subscription:
 
     ```bash
-    oc -n aws-load-balancer-operator patch subscription aws-load-balancer-operator --type='merge' -p '{"spec":{"config":{"volumes":[{"name":"trusted-ca","configMap":{"name":"trusted-ca"}}],"volumeMounts":[{"name":"trusted-ca","mountPath":"/etc/pki/tls/certs/albo-tls-ca-bundle.crt","subPath":"ca-bundle.crt"}]}}}'
+    oc -n aws-load-balancer-operator patch subscription aws-load-balancer-operator --type='merge' -p '{"spec":{"config":{"env":[{"name":"TRUSTED_CA_CONFIGMAP_NAME","value":"trusted-ca"}],"volumes":[{"name":"trusted-ca","configMap":{"name":"trusted-ca"}}],"volumeMounts":[{"name":"trusted-ca","mountPath":"/etc/pki/tls/certs/albo-tls-ca-bundle.crt","subPath":"ca-bundle.crt"}]}}}'
     ```
 
 3. Wait for the operator deployment to finish the rollout and verify that CA bundle is added:
 
     ```bash
-    oc -n aws-load-balancer-operator exec deploy/aws-load-balancer-operator-controller-manager -c manager -- ls -l /etc/pki/tls/certs/albo-tls-ca-bundle.crt
+    oc -n aws-load-balancer-operator exec deploy/aws-load-balancer-operator-controller-manager -c manager -- bash -c "ls -l /etc/pki/tls/certs/albo-tls-ca-bundle.crt; printenv TRUSTED_CA_CONFIGMAP_NAME"
 
     -rw-r--r--. 1 root 1000690000 5875 Jan 11 12:25 /etc/pki/tls/certs/albo-tls-ca-bundle.crt
+    trusted-ca
     ```
 
 4. _Optional_: make sure the operator is restarted every time the configmap contents change:
--- a/pkg/controllers/awsloadbalancercontroller/controller.go
+++ b/pkg/controllers/awsloadbalancercontroller/controller.go
@@ -19,7 +19,6 @@ package awsloadbalancercontroller
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	arv1 "k8s.io/api/admissionregistration/v1"
@@ -35,9 +34,11 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	albo "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
 	"github.com/openshift/aws-load-balancer-operator/pkg/aws"
@@ -59,19 +60,21 @@ const (
 // AWSLoadBalancerControllerReconciler reconciles a AWSLoadBalancerController object
 type AWSLoadBalancerControllerReconciler struct {
 	client.Client
-	Scheme      *runtime.Scheme
-	Namespace   string
-	Image       string
-	EC2Client   aws.EC2Client
-	ClusterName string
-	VPCID       string
-	AWSRegion   string
+	Scheme                 *runtime.Scheme
+	Namespace              string
+	Image                  string
+	EC2Client              aws.EC2Client
+	ClusterName            string
+	VPCID                  string
+	AWSRegion              string
+	TrustedCAConfigMapName string
 }
 
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=networking.olm.openshift.io,resources=awsloadbalancercontrollers/finalizers,verbs=update
 //+kubebuilder:rbac:groups="",resources=services;secrets,namespace=system,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=configmaps,namespace=system,verbs=get;list;watch
 //+kubebuilder:rbac:groups="networking.k8s.io",resources=ingressclasses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="config.openshift.io",resources=infrastructures,verbs=get;list;watch
 //+kubebuilder:rbac:groups="apps",resources=deployments,namespace=system,verbs=get;list;watch;create;update;patch;delete
@@ -162,6 +165,19 @@ func (r *AWSLoadBalancerControllerReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{RequeueAfter: secretMissingReEnqueueDuration}, nil
 	}
 
+	var trustCAConfigMap *corev1.ConfigMap
+	if r.TrustedCAConfigMapName != "" {
+		configMap, configMapExists, err := r.getConfigMap(ctx, r.TrustedCAConfigMapName, r.Namespace)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("failed to get the trusted CA configmap: %w", err)
+		}
+		if !configMapExists {
+			logger.Info("(Retrying) trusted CA config map doesn't exist", "configmap", r.TrustedCAConfigMapName)
+			return reconcile.Result{RequeueAfter: secretMissingReEnqueueDuration}, nil
+		}
+		trustCAConfigMap = configMap
+	}
+
 	sa, err := r.ensureControllerServiceAccount(ctx, r.Namespace, lbController)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to ensure AWSLoadBalancerController %q service account: %w", req.Name, err)
@@ -172,7 +188,7 @@ func (r *AWSLoadBalancerControllerReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, fmt.Errorf("failed to ensure ClusterRole and Binding for AWSLoadBalancerController %q: %w", req.Name, err)
 	}
 
-	deployment, err := r.ensureDeployment(ctx, r.Namespace, r.Image, sa, credSecretNsName.Name, servingSecretName, lbController)
+	deployment, err := r.ensureDeployment(ctx, sa, credSecretNsName.Name, servingSecretName, lbController, trustCAConfigMap)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to ensure Deployment for AWSLoadbalancerController %q: %w", req.Name, err)
 	}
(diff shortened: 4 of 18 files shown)


alebedev87:
Certified operator catalog index image not available:
```
Back-off pulling image "registry.redhat.io/redhat/certified-operator-index:v4.12"
```

/retest

gcs278 (commented):
Looks good overall, just some nit picks.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
     capabilities: Basic Install
     olm.skipRange: <0.2.0
     operatorframework.io/suggested-namespace: aws-load-balancer-operator
+    operators.openshift.io/infrastructure-features: '["proxy-aware"]'
```
gcs278, line 9: Could you help me understand what this does?

docs/proxy.md
```diff
@@ -0,0 +1,28 @@
...
+Those variables are then propagated down to the managed controller by AWS Load Balancer Operator.
+
+## Trusted Certificate Authority
+
+### Running operator
+Follow the instructions below to let AWS Load Balancer Operator trust a custom Certificate Authority (CA). The operator's OLM subscription has to be already created.
+
+1. Create the configmap containing the CA bundle in `aws-load-balancer-operator` namespace.
+    ```bash
+    oc -n aws-load-balancer-operator create configmap trusted-ca
```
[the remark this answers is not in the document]
alebedev87: Actually this configmap will be populated by cluster-network-operator thanks to[ the annotation](https://docs.openshift.com/container-platform/4.12/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki) from the command below.

config/manager/manager.yaml
```diff
@@ -37,6 +37,7 @@ spec:
         - "--leader-elect"
         - "--image=$(RELATED_IMAGE_CONTROLLER)"
         - "--namespace=$(TARGET_NAMESPACE)"
+        - "--trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)"
```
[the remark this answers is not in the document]
alebedev87: Yes, it'll be empty. So it'll be passed empty to the `AWSLoadBalancerControllerReconciler` which would mean to not add the additional volume and so on.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: It's a tag which informs the user of the operator that this operator is proxy aware ([doc](https://docs.openshift.com/container-platform/4.12/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-manual-annotations_osdk-generating-csvs)). 

The user will be able to filter operators by the presence of this feature:
image

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -152,149 +153,6 @@ func TestDesiredArgs(t *testing.T) {
...
-		Name:            b.name,
-		Image:           b.image,
-		Args:            b.args,
-		Env:             b.env,
-		VolumeMounts:    b.volumeMounts,
-		SecurityContext: b.securityContext,
-	}
-}
-
 func TestUpdateDeployment(t *testing.T) {
```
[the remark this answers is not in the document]
alebedev87: Added the annotation change test cases to `TestUpdateDeployment`.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
gcs278: Oh I see, that's neat!

docs/proxy.md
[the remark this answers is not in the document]
gcs278: Okay I see that now. Is the ca configmap function something that would be obvious to readers of this doc? Or is it worth a mention here?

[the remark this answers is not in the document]
alebedev87: Sorry, I'm not sure I understood the questions.

The configmap will be populated with the trusted CA bundle once the command from this doc is run:
```
oc -n aws-load-balancer-operator label cm trusted-ca config.openshift.io/inject-trusted-cabundle=true
```

The new flag [trusted-ca-configmap ](https://github.com/openshift/aws-load-balancer-operator/blob/1e8c7d2ed273579efd7ae458aed60047c0908a68/main.go#L92)mentions which format is expected from the configmap. That's exactly the format injected into the configmap annotated with `config.openshift.io/inject-trusted-cabundle` by the cluster-network-operator.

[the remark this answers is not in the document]
gcs278: > Actually this configmap will be populated by cluster-network-operator thanks to[ the annotation](https://docs.openshift.com/container-platform/4.12/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki) from the command below.

Sorry, I meant to say, it is worth mentioning something like this^ in these instructions? In other words, describing what the network operator is going to do in a little more detail.

```
1. Create an empty configmap containing the CA bundle in `aws-load-balancer-operator` namespace, then add the `config.openshift.io/inject-trusted-cabundle` annotation so the cluster-network-operator populates the configmap automatically.  
```

As a new reader, I was a little confused, I didn't understand where the `inject-trusted-cabundle` was going to inject, hence the suggestion, but I can understand if this something obvious for the average user of ALBO (or openshift for that matter).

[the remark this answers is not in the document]
alebedev87: Updated the doc, let me know if this is clearer now.

gcs278:
Thanks!
/lgtm
/hold
For @candita to take a look.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -206,10 +222,22 @@ func (r *AWSLoadBalancerControllerReconciler) getAWSLoadBalancerController(ctx c
...
+	}
+	return &cm, true, nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *AWSLoadBalancerControllerReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&albo.AWSLoadBalancerController{}, builder.WithPredicates(reconcileClusterNamedResource())).
+	bldr := ctrl.NewControllerManagedBy(mgr).
+		For(&albo.AWSLoadBalancerController{}, builder.WithPredicates(predicate.NewPredicateFuncs(hasName(controllerName)))).
```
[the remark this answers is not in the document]
alebedev87: This concrete predicate is checked by the e2e test. However the predicates I added for the trusted CA configmap aren't checked anywhere. I'd prefer them to be checked in the e2e too but that would require a lot of effort - new CI job with a cluster wide proxy in the middle which has its own certificate and so on.
That's why I left it as is but after your comment I gave it another thought and found that ALBO already has `envtest` set up for the unit test. So I added the unit test for the watch predicates, see the last commit.

pkg/controllers/suite_test.go
```diff
@@ -17,19 +17,26 @@ limitations under the License.
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
+	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+
 	networkingolmv1alpha1 "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
+	albc "github.com/openshift/aws-load-balancer-operator/pkg/controllers/awsloadbalancercontroller"
```
candita, line 39: We usually have openshift imports after go package imports, followed by an alphabetized list of k8s.io and other imports.

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -218,20 +246,43 @@ func (r *AWSLoadBalancerControllerReconciler) SetupWithManager(mgr ctrl.Manager)
 		Owns(&appsv1.Deployment{}).
 		Owns(&corev1.Service{}).
 		Owns(&arv1.ValidatingWebhookConfiguration{}).
-		Owns(&arv1.MutatingWebhookConfiguration{}).
-		Complete(r)
+		Owns(&arv1.MutatingWebhookConfiguration{})
```
candita, line 249: You removed the call to `Complete` here.  What does the work that `Complete` used to do (actual building)?

pkg/controllers/awsloadbalancercontroller/controller.go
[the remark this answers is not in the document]
candita: Let's step back a bit.  Does `predicate.NewPredicateFuncs(hasName(controllerName))` represent the same list as `reconcileClusterNamedResource()` used to represent?  If not, why not?

pkg/controllers/awsloadbalancercontroller/controller.go
```diff
@@ -218,20 +246,43 @@ func (r *AWSLoadBalancerControllerReconciler) SetupWithManager(mgr ctrl.Manager)
...
+				{
+					NamespacedName: types.NamespacedName{
+						Name: controllerName,
+					},
+				},
+			}
+		}
+
+		bldr = bldr.Watches(&source.Kind{Type: &corev1.ConfigMap{}},
+			handler.EnqueueRequestsFromMapFunc(clusterALBCInstance),
```
candita, line 263: I need some help on this, will come back to it.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -15,16 +17,20 @@ import (
...
 const (
-	appLabelName    = "app.kubernetes.io/name"
-	appName         = "aws-load-balancer-operator"
-	appInstanceName = "app.kubernetes.io/instance"
+	appLabelName        = "app.kubernetes.io/name"
+	appName             = "aws-load-balancer-operator"
+	appInstanceName     = "app.kubernetes.io/instance"
+	trustedCAAnnotation = "networking.olm.openshift.io/trusted-ca-configmap-hash"
+	// controllerContainerName is the name of the AWS load balancer controller's container.
+	controllerContainerName = "controller"
```
candita, line 33: Please be more specific in the name.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -46,22 +52,40 @@ const (
...
-	exists, current, err := r.currentDeployment(ctx, deploymentName, namespace)
+	trustCAConfigMapName, trustCAConfigMapHash := "", ""
+	if trustCAConfigMap != nil {
+		trustCAConfigMapName = trustCAConfigMap.Name
+		configMapHash, err := buildMapHash(trustCAConfigMap.Data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build the trusted CA configmap's hash: %w", err)
+		}
+		trustCAConfigMapHash = configMapHash
+	}
```
candita, line 81: Move this section down to after L86.  In other words, check first to make sure the deployment exists before you do this work.

pkg/controllers/awsloadbalancercontroller/controller.go
candita: How about adding some comments here?

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -130,7 +154,7 @@ func desiredDeployment(name, namespace, image, vpcID, clusterName, awsRegion, cr
 									Name:  awsSDKLoadConfigName,
 									Value: "1",
 								},
-							},
+							}, proxy.ReadProxyVarsFromEnv()...),
```
candita, line 157: Comments here too.  I'm not sure why you would want to take all the proxy var from env for the container.

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -200,6 +224,30 @@ func desiredDeployment(name, namespace, image, vpcID, clusterName, awsRegion, cr
 	if controller.Spec.Config != nil && controller.Spec.Config.Replicas != 0 {
 		d.Spec.Replicas = pointer.Int32(controller.Spec.Config.Replicas)
 	}
+	if trustedCAConfigMapName != "" {
```
candita, line 227: Why don't you check for trustedCAVolumeName too?

pkg/controllers/awsloadbalancercontroller/deployment.go
```diff
@@ -15,16 +17,20 @@ import (
...
 )
 
 const (
-	appLabelName    = "app.kubernetes.io/name"
-	appName         = "aws-load-balancer-operator"
-	appInstanceName = "app.kubernetes.io/instance"
+	appLabelName        = "app.kubernetes.io/name"
+	appName             = "aws-load-balancer-operator"
+	appInstanceName     = "app.kubernetes.io/instance"
+	trustedCAAnnotation = "networking.olm.openshift.io/trusted-ca-configmap-hash"
```
candita, line 31: Can you make this work without adding an annotation?  If not, why not?

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -152,149 +153,6 @@ func TestDesiredArgs(t *testing.T) {
 	}
 }
 
-type testDeploymentBuilder struct {
```
candita, line 155: Be kind to your reviewer/s, don't move 150 lines of code to another place.  Now we have to review all the code you moved, to make sure it didn't change.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
```diff
@@ -686,17 +585,239 @@ func TestEnsureDeployment(t *testing.T) {
...
+							Path:              "token",
+						},
+					}},
+				}}},
+				corev1.Volume{Name: "trusted-ca", VolumeSource: corev1.VolumeSource{ConfigMap: &corev1.ConfigMapVolumeSource{LocalObjectReference: corev1.LocalObjectReference{Name: "test-trusted-ca"}}}},
+			).build(),
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
+			for k, v := range tc.existingEnvVars {
```
candita, line 798: This only runs for one test case.  In other words, only one of your test cases contains `existingEnvVars`.  Seems like it would be better in its own test instead of here.

candita:
Small changes requested, otherwise:
/lgtm

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: No. The annotation is needed in the template pod spec of the managed controller to re-trigger the deployment in case the contents of the trusted CA configmap changed. The annotation contains the hash of the trusted CA configmap: changed configmap contents => changed hash => changed template podspec => redeployment of the aws load balancer controller.

pkg/controllers/test/crd/credentialsrequest.yaml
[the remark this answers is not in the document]
alebedev87: I had to add the CCO's CredentialsRequest CRD to be able [to test all the predicates](https://github.com/openshift/aws-load-balancer-operator/pull/90#discussion_r1108950639) of the operator's watches. The operator watches for `CredentialsRequest` custom resource, it's OpenShift specific therefore not present in `envtest` by default.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
alebedev87: Done.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: Done.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: Done.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: Done.

alebedev87:
/test e2e-aws-rosa-operator

`Step e2e-aws-rosa-operator-rosa-operators-wait-ready failed after`. One of the cluster operators didn't become ready in the given time.

pkg/controllers/suite_test.go
alebedev87: Reformatted the import according to [this goio example](https://github.com/go-imports-organizer/goio/blob/main/examples/openshift_kubernetes.yaml) ([slack thread](https://redhat-internal.slack.com/archives/C0AL30LMR/p1678407115379399)).

alebedev87:
/test e2e-aws-rosa-operator

Cluster creation failed.

gcs278:
Reviewed the updates.
/lgtm

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: The volume is added by the operator, the user doesn't have a means to influence the volume. Also the same configmap [is used by the operator](https://github.com/openshift/aws-load-balancer-operator/blob/main/docs/proxy.md#running-operator) to talk to AWS at the staartup, so the operator would fail to start before we reach this point in code.

pkg/controllers/awsloadbalancercontroller/controller.go
candita: Shouldn't you check for changes first before you requeue it?

pkg/controllers/awsloadbalancercontroller/deployment.go
candita: We use a different idiom to compute changes, everywhere else in OpenShift.  We compare the previous and new, and don't need to add annotations.  I would prefer if you follow that idiom.

pkg/controllers/awsloadbalancercontroller/deployment.go
candita: So you just take everything, not filtering for specific env vars that you need.

alebedev87: No, [ReadProxyVarsFromEnv](https://pkg.go.dev/github.com/operator-framework/operator-lib@v0.11.0/proxy#ReadProxyVarsFromEnv) takes only the proxy vars as the name suggests.

pkg/controllers/awsloadbalancercontroller/controller.go
alebedev87: What changes do you have in mind?

The logic here is to re-trigger the awsloadbalancercontroller's reconciliation loop when the controller runtime sees any changes on the trusted CA configmap we specified in the predicates. The reconciliation loop aligns the operands with the current state. The trusted CA configmap is a part of the alignment - we calculate its checksum and see if it needs a new rollout of the managed deployment.

alebedev87: Update: I read [this comment](https://github.com/openshift/aws-load-balancer-operator/pull/90#discussion_r1144992436) and I guess I know where the confusion comes from. The trusted CA configmap on its own is not part of the desired state, but the consumption of this configmap is. The operator has to make sure a volume and volumeMount is added to the managed aws lb controller's deployment. The contents of the configmap though are opaque to the operator, it makes the managed deployment consume whatever is in there.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: I see what you mean. Yes, all the operators compare the new and the desired states. The difference here is that the trusted CA configmap is not part of the desired state. The operator takes it as an opaque object. However the operator's job is to ensure this configmap is consumed by the managed aws lb controller's deployment. That is done by having the volume and the volumeMount as a part of the desired state (`pkg/controllers/awsloadbalancercontroller/deployment.go` line 231).
The annotation though is just a way to let the aws-load-balancer-controller reload CA certs from the host (container). The annotation would not be necessary if the aws-load-balancer-controller could [fsnotify](https://pkg.go.dev/github.com/fsnotify/fsnotify) on the system certs.

alebedev87:
The push above only modifies `docs/proxy.md`. I noticed that it was wrongly rebased after the merge of https://github.com/openshift/aws-load-balancer-operator/pull/91. I forgot to add the parts for the controller, corrected now.

pkg/controllers/suite_test.go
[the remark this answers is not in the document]
alebedev87: I thought we were ok with the format proposed in goio ([comment from before](https://github.com/go-imports-organizer/goio/blob/main/README.md)). According to goio's example - the module imports go last as I did from the very beginning.

pkg/controllers/awsloadbalancercontroller/deployment_test.go
alebedev87: Sorry about that, will try to minimize such moves.

lihongan:
@gcs278 @candita   It this OK to cancel `hold` label now ? seems all comments are addressed

pkg/controllers/awsloadbalancercontroller/controller.go
candita: I'm going to ask someone else to review this.

pkg/controllers/suite_test.go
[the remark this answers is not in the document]
candita: I'm not ok with it.

pkg/controllers/awsloadbalancercontroller/deployment.go
Miciah: @alebedev87 , can you add godoc to `trustedCAAnnotation` to explain that its purpose is to force the controller to restart in order to reload the trusted CA when it changes?

candita:
> @gcs278 @candita It this OK to cancel `hold` label now ? seems all comments are addressed

@gcs278 I'll leave it up to you, since you added the `hold` label.

pkg/controllers/awsloadbalancercontroller/deployment.go
alebedev87: @Miciah : comment added. Also the comment in `controller.go#BuildManagedController` was updated  to explain better what happens after the reconciliation gets triggered from the trusted CA configmap events.

alebedev87:
Addressed [the comment](https://github.com/openshift/aws-load-balancer-operator/pull/90#discussion_r1156227029).

alebedev87:
`e2e-aws-rosa-operator` failed due to a change in `rosa describe cluster` output. [This PR](https://github.com/openshift/release/pull/38023) addresses the change.

gcs278:
Re-reviewed.
/lgtm

gcs278:
I don't think my hold applies anymore
/hold cancel

PR #92, opened 2023-02-28
NE-1225: adapt e2e tests to enable ROSA platform with STS service

alebedev87:
This PR adapts the existing e2e tests for t[he new CI job](https://github.com/openshift/release/pull/36655) which runs on a ROSA STS cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,18 +16,23 @@ describes the design and implementation of the operator in more detail.
    1. [STS Clusters](/docs/install.md#sts-clusters)
 3. [Tutorial](/docs/tutorial.md)
 4. [Local Development](#local-development)
+    1. [Build the operand image](#build-the-operand-image)
+    2. [Running the operator](#running-the-operator)
+    3. [Running the end-to-end tests](#running-the-end-to-end-tests)
+    4. [Running the end-to-end tests on ROSA STS cluster](#running-the-end-to-end-tests-on-rosa-sts-cluster)
 
 ## Local Development
 
 ### Build the operand image
 
-The operand image must be built first. Our fork of the operand
-is [here](https://github.com/openshift/aws-load-balancer-controller/). Clone 
-the repository and build the image and push it to a registry which is 
-accessible from the test cluster.
+**Note**: only needed for unmerged changes, all merged changes get published in a public quay.io repository
+
+The operand image must be built first. Clone [the OpenShift fork of the operand](https://github.com/openshift/aws-load-balancer-controller),
+build the image and push it to a registry which is accessible from the test cluster.
 
 ```bash
 git clone https://github.com/openshift/aws-load-balancer-controller.git
+cd aws-load-balancer-controller
 IMG=quay.io/$USER/aws-load-balancer-controller
 podman build -t $IMG -f Dockerfile.openshift
 podman push $IMG
@@ -38,7 +43,7 @@ podman push $IMG
 1. Replace the operand image in the file `config/manager/manager.yaml` in 
    the environment variable `RELATED_IMAGE_CONTROLLER` with the image 
    created in the previous step.
-2. Build and push the image to an image registry.
+2. Build and push the operator image to an image registry.
     ```bash
     export IMG=quay.io/$USER/aws-load-balancer-operator:latest
     make image-build image-push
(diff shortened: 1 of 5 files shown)


alebedev87:
`TestAWSLoadBalancerControllerWithCredentialsSecret` test expectedly fails because the creation of the controller secret was removed. The [CI PR](https://github.com/openshift/release/pull/36655) will fix this as it now adds the controller secret for the non ROSA job too.

README.md
```diff
@@ -75,3 +80,22 @@ tests with the following command:
...
+```bash
+aws wafv2 create-web-acl --name "echoserver-acl" --scope REGIONAL --default-action '{"Block":{}}'  --visibility-config '{"MetricName":"echoserver","CloudWatchMetricsEnabled": false,"SampledRequestsEnabled":false}'
+aws waf-regional create-web-acl --name "echoserverclassicacl" --metric-name "echoserverclassicacl" --default-action '{"Type":"BLOCK"}' --change-token "$(aws waf-regional get-change-token)"
+```
+
+Now you can run the e2e test with the following commands:
+```bash
+export ALBO_E2E_PLATFORM=ROSA
+export ALBO_E2E_WAF_WEBACL_ID=<wafregional-webacl-id>
+export ALBO_E2E_WAFV2_WEBACL_ARN=<wafv2-webacl-arn>
```
candita, line 99: Do you need to describe how to get the `wafregional-webacl-id` and `wafv2-webacl-arn`?  If they come from the two commands above, maybe mention that above.

test/e2e/operator_test.go
```diff
@@ -151,7 +162,12 @@ func TestAWSLoadBalancerControllerWithDefaultIngressClass(t *testing.T) {
 	t.Log("Creating aws load balancer controller instance with default ingress class")
 
 	name := types.NamespacedName{Name: "cluster", Namespace: "aws-load-balancer-operator"}
-	alb := newALBCBuilder().withName(name).build()
+
+	bldr := newALBCBuilder().withName(name)
+	if isOnROSA() {
+		bldr.withCredSecret(controllerSecretName)
+	}
+	alb := bldr.build()
```
candita, line 170: Should probably make a helper function for this, it is used 4 other times.

test/e2e/operator_test.go
```diff
@@ -514,53 +517,63 @@ func TestAWSLoadBalancerControllerWithWAFv2(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("failed to create aws wafv2 acl due to %v", err)
+			}
+			acl = createdACL.Summary
 		}
-		acl = createdACL.Summary
-	}
 
-	t.Logf("Got AWS WAFv2 ACL. ID: %s, Name: %s", *acl.Id, *acl.Name)
+		aclARN = *acl.ARN
```
candita, line 546: Can acl ever be nil?

test/e2e/operator_test.go
```diff
@@ -624,45 +641,56 @@ func TestAWSLoadBalancerControllerWithWAFRegional(t *testing.T) {
...
+			MetricName:    aws.String("echoserverclassicacl"),
+			Name:          aws.String("echoserverclassicacl"),
+			ChangeToken:   token.ChangeToken,
 		})
 		if err != nil {
-			t.Fatalf("failed to delete aws waf regional acl due to %v", err)
+			t.Fatalf("failed to create aws waf regional acl due to %v", err)
 		}
-	}()
+		webACLID = *acl.WebACL.WebACLId
```
candita, line 661: Could acl ever be nil?

test/e2e/operator_test.go
```diff
@@ -883,6 +869,52 @@ func TestIngressGroup(t *testing.T) {
 	}
 }
 
+func ensureCredentialsRequest() error {
```
candita, line 872:
```suggestion
func ensureCredentialsRequest() error {
    t.Helper()
```

test/e2e/operator_test.go
```diff
@@ -905,3 +937,7 @@ func createTestWorkload(t *testing.T, namespace string) (*corev1.Service, *corev
 
 	return echosvc, ns
 }
+
+func isOnROSA() bool {
```
candita, line 941:
```suggestion
func isOnROSA() bool {
    t.Helper()
```

test/e2e/util.go
```diff
@@ -20,16 +20,13 @@ import (
...
 
-	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
-
 	"github.com/aws/aws-sdk-go-v2/service/wafv2"
 	wafv2types "github.com/aws/aws-sdk-go-v2/service/wafv2/types"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	albo "github.com/openshift/aws-load-balancer-operator/api/v1alpha1"
```
candita, line 29: nit: Normally openshift imports come before k8s and others, and the others are alphabetized.

candita:
Just a few nits, otherwise:
/lgtm

test/e2e/operator_test.go
alebedev87: No. We try to find it first and if it doesn't exist we try to create it. If the creation fails we fail the test.

test/e2e/operator_test.go
alebedev87: No, [wafregional.CreateWebACL method](https://github.com/aws/aws-sdk-go-v2/blob/service/wafregional/v1.13.6/service/wafregional/api_op_CreateWebACL.go#L53) returns `nil` only if there is an error. We fail the test in case the error is returned.

test/e2e/operator_test.go
alebedev87: `Helper()` makes sense only in the functions which do something with the test: fail it, log something.  The helper functions should have `t *testing.T` as one of the parameters.

However, I found other functions which should be helpers: `createTestWorkload` and `waitForHTTPClientCondition` (from `util.go`).

test/e2e/operator_test.go
alebedev87: Same answer as in [this comment](https://github.com/openshift/aws-load-balancer-operator/pull/92#discussion_r1138447991).

test/e2e/operator_test.go
alebedev87: Done.

test/e2e/util.go
candita: It is an OpenShift style thing, but you have to admit the alphabetized list is super helpful.

PR #94, opened 2023-06-07
OCPBUGS-13973: bump k8s modules to 0.27.2

alebedev87:
Bump of `k8s.io/api@0.27.2`. `controller-runtime@0.15.0` bump was mandatory as it's the first version supporting k8s api >=0.27.x.

Changes related to `controller-runtime@0.15.0`:
- Manager options changed:
  - `NewClient` changed the signature and the cache handling
  - `Namespace` is deprecated now, `Cache` field was used instead
- `WithStatusSubresource` method has to be used in the fake client if the status is expected to be updated
- `source.Kind` is not public anymore
- `sigs.k8s.io/controller-runtime/pkg/envtest/printer` package has been removed

Also `golangci-lint` (part of `verify` job) forced the following changes:
- `k8s.io/apimachinery/pkg/util/sets` deprecated type specific sets in favor of a generic `Set`
- `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait` deprecated a lot of `PollX` methods in favor of the context aware alternatives.
- `k8s.io/utils/pointer` deprecated all the methods with `Ptr` in the name in favor of the methods without it


test/e2e/util.go
```diff
@@ -39,12 +39,12 @@ var (
 	allowPrivilegeEscalation = false
 )
 
-func waitForDeploymentStatusCondition(t *testing.T, cl client.Client, timeout time.Duration, deploymentName types.NamespacedName, conditions ...appsv1.DeploymentCondition) error {
+func waitForDeploymentStatusCondition(ctx context.Context, t *testing.T, cl client.Client, timeout time.Duration, deploymentName types.NamespacedName, conditions ...appsv1.DeploymentCondition) error {
 	t.Helper()
 
-	return wait.PollImmediate(10*time.Second, timeout, func() (bool, error) {
+	return wait.PollUntilContextTimeout(ctx, 10*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
```
gcs278, line 45: What's the motivation to  use `PollUntilContextTimeout`. Is this a bit like a hard-stop when `cl.Get` get stuck?

alebedev87: `PollImmediate` [was deprecated](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate) in favor of `PollUntilContextTimeout`. The main advantage I see is the new parameter - `context`. It's passed down to the polling function and checked in between the calls. So, this makes `PollUntilContextTimeout` responsive to the external interruption (SIGTERM or similar) unlike `PollImmediate` which has to get an error or time out.

Unfortunately we don't really use this new helpful functionality in the e2e tests. We always pass `context.TODO()` due to a lack of an external interruption (pity that golang's `testing` doesn't pass `context` to react to SIGINT or `go test` timeout).  For instance, the operator binary [does have a signal handler](https://github.com/openshift/aws-load-balancer-operator/blob/main/main.go#L173) which can cancel the context if the process is interrupted or terminated.

gcs278: Ack thanks - I didn't realize it was deprecated - that's good to know because we use that everywhere.

PR #95, opened 2023-06-23
OCPBUGS-14631: use openshift docs for operatorhub prerequsiste instructions

alebedev87:
The instructions for the STS mode come from the openshift docs now, not upstream.

### DIFF
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -101,61 +101,68 @@ spec:
       name: targetgroupbindings.elbv2.k8s.aws
       version: v1beta1
   description: |-
-    Operator to simplify management of aws-load-balancer-controller
+    Operator to simplify the management of aws-load-balancer-controller.
 
     ### Prerequisites for installation
     The operator requires AWS credentials in the installation namespace before it can be installed.
 
+    #### Instructions for non-STS cluster
     1. Create the operator namespace with the following command:
-    ```bash
-    oc create namespace aws-load-balancer-operator
-    ```
-    2. Create the `CredentialsRequest` in the `openshift-cloud-credential-operator`
-       namespace with the following command:
 
-    ```
-    cat <<EOF | oc create -f -
-    apiVersion: cloudcredential.openshift.io/v1
-    kind: CredentialsRequest
-    metadata:
-      name: aws-load-balancer-operator
-      namespace: openshift-cloud-credential-operator
-    spec:
-      providerSpec:
-        apiVersion: cloudcredential.openshift.io/v1
-        kind: AWSProviderSpec
-        statementEntries:
-          - action:
-              - ec2:DescribeSubnets
-            effect: Allow
-            resource: "*"
-          - action:
-              - ec2:CreateTags
-              - ec2:DeleteTags
-            effect: Allow
-            resource: arn:aws:ec2:*:*:subnet/*
-          - action:
-              - ec2:DescribeVpcs
-            effect: Allow
-            resource: "*"
-      secretRef:
+      ```bash
+      oc create namespace aws-load-balancer-operator
+      ```
+
+    2. Create the `CredentialsRequest` in the `openshift-cloud-credential-operator` namespace with the following command:
+
+      ```bash
+      cat <<EOF | oc create -f -
+      apiVersion: cloudcredential.openshift.io/v1
+      kind: CredentialsRequest
+      metadata:
         name: aws-load-balancer-operator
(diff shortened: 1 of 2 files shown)


bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -101,61 +101,68 @@ spec:
...
+      oc get secret -n aws-load-balancer-operator aws-load-balancer-operator
+      ```
 
     After this the operator can be installated through the console or through the command line.
 
-    __Note about UPI:__ For UPI based installs, additional documentation can be found in
-    [VPC and subnets](https://github.com/openshift/aws-load-balancer-operator/blob/main/docs/prerequisites.md#vpc-and-subnets)
+    #### Instructions for STS cluster (inclusing ROSA in STS mode)
+
+    [Bootstrapping AWS Load Balancer Operator on Security Token Service cluster](https://docs.openshift.com/container-platform/4.13/networking/aws_load_balancer_operator/installing-albo-sts-cluster.html#nw-bootstra-albo-on-sts-cluster_albo-sts-cluster).
```
gcs278, line 161: TIL you can use `latest` in docs link to reference latest version. I feel like that's better than hardcoding 4.13 here. OSSM also uses [latest](https://github.com/maistra/istio-operator/blob/maistra-2.4/manifests-servicemesh/2.4.3/servicemeshoperator.v2.4.3.clusterserviceversion.yaml#L149) in their links. Let me know if you disagree.
```suggestion
    [Bootstrapping AWS Load Balancer Operator on Security Token Service cluster](https://docs.openshift.com/container-platform/latest/networking/aws_load_balancer_operator/installing-albo-sts-cluster.html#nw-bootstra-albo-on-sts-cluster_albo-sts-cluster).
```

gcs278 (commented):
Just one comment, everything else looks good.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: I agree that `latest` makes sense for `main` branch. We will have to remember to change the link back to the one with "OCP version pinned" when we do a new release (create a new release branch). Just noticed that we already made a similar mistake for the UPI based instructions: [release-0.2 links](https://github.com/openshift/aws-load-balancer-operator/blob/release-0.2/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml#L73) `main` branch documentation.

alebedev87: > Just noticed that we already made a similar mistake for the UPI based instructions: [release-0.2 links](https://github.com/openshift/aws-load-balancer-operator/blob/release-0.2/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml#L73) main branch documentation.

@gcs278 : can you help me to fix the mistake?

- https://github.com/openshift/aws-load-balancer-operator/pull/100
- https://github.com/openshift/aws-load-balancer-operator/pull/101
- https://github.com/openshift/aws-load-balancer-operator/pull/102

alebedev87: Link updated to use `latest`.

gcs278: > We will have to remember to change the link back to the one with "OCP version pinned" when we do a new release (create a new release branch).

Sure, but I wonder if this process will be forgotten in the future. I wonder if there a better way to do this, like using a version file with mapping of ALBO to OCP releases, and using that version file to generate docs with the right links. Probably overkill for now, just thinking out loud.

Either way, we can proceed with remembering to adjust it later.

gcs278:
@alebedev87 whoops, I'm not in the owner's file. You can approve yourself.

alebedev87:
@gcs278 : indeed, I fixed this with https://github.com/openshift/aws-load-balancer-operator/pull/103.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: > Sure, but I wonder if this process will be forgotten in the future. 

Yes, there is a high chance indeed.

> I wonder if there a better way to do this, like using a version file with mapping of ALBO to OCP releases, and using that version file to generate docs with the right links. Probably overkill for now, just thinking out loud.

We can think of it. This repository misses such a mapping ([similar to ExternalDNS](https://github.com/openshift/external-dns-operator/blob/main/docs/openshift.md#release-branch-mapping)).

PR #100, opened 2023-08-21
docs: pin the prerequisite link to release-0.1

alebedev87:
The documentation on released versions should be pinned to the specific version to avoid the instructions from the future releases.


gcs278:
Seems acceptable for now. Just curious if there is a more elegant way to manage these links.
/lgtm
/approve

PR #109, opened 2023-08-30
NE-1342: self provision with credentials using CredentialsRequest

alebedev87:
With this PR, ALBO can provision itself with AWS credentials using CCO's `CredentialsRequest`.

**Before**: the user had to do a mandatory pre-installation manual step to create the credentials secret either via applying `CredentialsRequest` (non-STS clusters) or via a manual secret creation (STS clusters). The operator consumed the provided secret as a `volumeMount` therefore preventing the operator from starting without this secret. The reason for the manual step was the fact that CCO didn't do the secret provisioning on STS clusters.

**Now**: the operator creates `CredentialsRequest` during startup. Then CCO provisions the secret on both non-STS and STS clusters. The operator waits until the secret is provisioned and creates aws credentials file which in turn consumed by AWS SDK. Some input from the user is still needed though, for STS clusters the user has to provide the ARN of the role. This approach is detailed in [the CCO EP](https://github.com/openshift/enhancements/pull/1339) and is now mandatory for all OLM operators which deal with AWS STS.

**Notable changes to the existing code**:
- IAM policy for the operator is now generated as code too (same as for the controller)
- `iamctl` binary modified to skip the compacting (`minify` in this repo's terms) as it's too generous and unnecessary for the operator's IAM policy (small enough to fit into `CredentialsRequest`)
- `pkg/controllers/awsloadbalancercontroller/utils/` moved to `pkg/utils/` in order to be able to use the `utils/test` package in `pkg/operator/`.
- credentials request update logic is put into a dedicated package in order to be used in `main.go` and in the controller: `pkg/utils/resource/update/`.


alebedev87:
> registry.build05.ci.openshift.org/ci/managed-clonerefs: received unexpected HTTP status: 504 Gateway Time-out

/test unit
/test verify
/test images
/test ci-index-aws-load-balancer-operator-bundle

alebedev87:
I forgot to update README.md and the manual instructions from CSV.

alebedev87:
Detailed operator installation instructions added. The operator may have problems to start w/o them.

pkg/operator/credentials.go
```diff
@@ -0,0 +1,159 @@
...
+func ProvisionCredentials(ctx context.Context, client client.Client, secretNamespace string) (string, error) {
+	roleARN := os.Getenv(roleARNEnvVar)
+	if roleARN != "" && !arn.IsARN(roleARN) {
+		return "", fmt.Errorf("provided role arn is invalid: %q", roleARN)
+	}
+
+	credReq := buildCredentialsRequest(secretNamespace, operatorCredentialsSecretName, roleARN)
+
+	// create credentials request resource
+	if err := client.Create(ctx, credReq); err != nil {
```
gcs278, line 60: I ran into an issue. When testing this PR with AWS STS, I forgot to put the the `ROLEARN` in the manager deployment yaml as a ENV variable. I recreated the deployment with `ROLEARN`, but scratched my head why it still wasn't working.

Then I realized it doesn't compare the content of the credential request, but just ensures an object exists with the same name. So I had to go delete the CR manually to get it to work.

Not saying it needs to compare, but do you think users would benefit from a comparison? Just curious if you considered these scenarios and if you are okay with the implications.

docs/prerequisites.md
```diff
@@ -132,28 +125,44 @@ If you prefer not to use `ccoctl`, or your system doesn't support it, the AWS CL
...
+      targetNamespaces: []
     EOF
-    ```
-    **Note**: mind the format of the credentials file, examples can be found in [OCP documentation](https://docs.openshift.com/container-platform/4.13/authentication/managing_cloud_provider_credentials/cco-mode-sts.html#sts-mode-about_cco-mode-sts).
-
-6. Create the operator's credentials secret with the generated aws credentials:
 
-    ```bash
-    oc -n aws-load-balancer-operator create secret generic aws-load-balancer-operator --from-file=credentials=albo-operator-aws-credentials.cfg
+    $ ROLEARN="<iam-role-arn>"
```
gcs278, line 147: Should this just be `$OPERATOR_ROLE_ARN`?

docs/prerequisites.md
```diff
@@ -132,28 +125,44 @@ If you prefer not to use `ccoctl`, or your system doesn't support it, the AWS CL
...
+      namespace: aws-load-balancer-operator
+    spec:
+      channel: stable-v1
+      name: aws-load-balancer-operator
+      source: redhat-operators
+      sourceNamespace: openshift-marketplace
+      config:
+        env:
+        - name: ROLEARN
+          value: "${ROLEARN}"
```
gcs278, line 163: Or conversely `$OPERATOR_ROLE_ARN`, which was created above.

gcs278 (commented):
Looks great, I just have a bunch of questions and minor nit picks. I spun up an STS cluster, tried out installing ALBO by creating iam role with `ccoctl` as well as deleting the IAM role and creating it via the AWS CLI. Everything worked well.

test/e2e/operator_test.go
gcs278, line 1: You may be getting to this later, but when I install ALBO with this PR (STS cluster) and try to run the E2E test, it indefinitely hangs. It tries to make another Credential Request, but since I already have one, it looks like it doesn't provision (create the E2E secret). 

Is STS supported for E2E testing yet? I think we need to provide the IAM Role ARN right?

docs/prerequisites.md
alebedev87: Yes, good catch.

docs/prerequisites.md
```diff
@@ -132,28 +125,44 @@ If you prefer not to use `ccoctl`, or your system doesn't support it, the AWS CL
...
+    $ ROLEARN="<iam-role-arn>"
+
+    $ cat <<EOF | oc apply -f -
+    apiVersion: operators.coreos.com/v1alpha1
+    kind: Subscription
+    metadata:
+      name: aws-load-balancer-operator
+      namespace: aws-load-balancer-operator
+    spec:
+      channel: stable-v1
```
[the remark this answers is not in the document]
alebedev87: > Would it be better to drop the channel so the default will be used?

Does it work? I don't remember seeing it documented. I always thought that the only reason for the default channel is to show which one to install in OperatorHub UI.

docs/prerequisites.md
```diff
@@ -1,93 +1,86 @@
...
-        - [Option 1. Using ccoctl](#option-1-using-ccoctl)
-        - [Option 2. Using the AWS CLI](#option-2-using-the-aws-cli)
+- [IAM Role for STS clusters](#iam-role-for-sts-cluster)
+    - [Option 1. Using ccoctl](#option-1-using-ccoctl)
+    - [Option 2. Using the AWS CLI](#option-2-using-the-aws-cli)
 - [VPC and Subnets](#vpc-and-subnets)
     - [VPC](#vpc)
     - [Subnets](#subnets)
         - [Public subnets](#public-subnets)
         - [Private subnets](#private-subnets)
```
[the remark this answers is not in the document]
alebedev87: Yeah, just may be placed right after the Pre-requisites chapter title.

alebedev87 (commented):
First run over the comments, easy things only.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -104,63 +104,13 @@ spec:
...
-    3. Ensure the credentials have been correctly provisioned:
-
-      ```bash
-      oc get secret -n aws-load-balancer-operator aws-load-balancer-operator
-      ```
-
-    After this the operator can be installated through the console or through the command line.
-
-    #### Instructions for STS cluster (inclusing ROSA in STS mode)
+    ### Instructions for STS cluster (inclusing ROSA in STS mode)
```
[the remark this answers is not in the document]
alebedev87: Changed to 4 dashes.

docs/prerequisites.md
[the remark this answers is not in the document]
alebedev87: Done.

alebedev87:
`main.go` updated to loop over the first AWS call to fix the permanently failing first start of the operator container.

docs/prerequisites.md
alebedev87: Done.

docs/prerequisites.md
```diff
@@ -1,93 +1,86 @@
...
 
-## AWS credentials
-Additional AWS credentials are needed for the operator to be successfully installed. This is needed to interact with subnets and VPCs.
+# IAM Role for STS clusters
```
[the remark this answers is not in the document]
alebedev87: Changed to 2 hashes.

Makefile
```diff
@@ -126,14 +126,18 @@ vet: ## Run go vet against code.
...
 	go fmt -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
 	go vet -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
+	$(IAMCTL_BINARY) -i $(IAMCTL_ASSETS_DIR)/operator-iam-policy.json -o ./pkg/operator/$(IAMCTL_OUTPUT_FILE) -p operator -n
+	go fmt -mod=vendor ./pkg/operator/$(IAMCTL_OUTPUT_FILE)
+	go vet -mod=vendor ./pkg/operator/$(IAMCTL_OUTPUT_FILE)
 
 .PHONY: iam-gen
 iam-gen:
 	./hack/generate-iam-from-credrequest.sh ./hack/operator-credentials-request.yaml ./hack/operator-permission-policy.json
+	cp ./hack/operator-permission-policy.json $(IAMCTL_ASSETS_DIR)/operator-iam-policy.json
```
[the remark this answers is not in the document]
alebedev87: The idea was to keep the single source of truth for the operator's permissions: `hack/operator-credentials-request.yaml`. It's mentioned in the dev docs (from this repository) and OCP docs as a way to create IAM role using `ccoctl`. It's also used in the CI steps.

I agree that it looks redundant. I just thought of making a smooth, incremental change which would stay consistent with the rest and doesn't break anything. We can think of removing the redundancy later.

docs/prerequisites.md
```diff
@@ -1,93 +1,86 @@
...
 - Using a pre-defined `CredentialsRequest`.
 - Using pre-defined AWS manifests.
 
 For handling `CredentialsRequests`, the cloud credential operator utility, [ccoctl](https://docs.openshift.com/container-platform/latest/authentication/managing_cloud_provider_credentials/cco-mode-sts.html#cco-ccoctl-configuring_cco-mode-sts), can be utilized.
 If you prefer not to use `ccoctl`, or your system doesn't support it, the AWS CLI can be an alternative.
 
-#### Option 1. Using `ccoctl`
+## Option 1. Using `ccoctl`
 
 1. [Extract and prepare the `ccoctl` binary](https://docs.openshift.com/container-platform/4.13/authentication/managing_cloud_provider_credentials/cco-mode-sts.html#cco-ccoctl-configuring_cco-mode-sts)
```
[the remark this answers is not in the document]
alebedev87: I agreed with you when we put `latest` in a link to the ALBO doc in the OperatorHub's installation page. Because it was our doc and `main` branch should always be consistent with the latest docs.
However here we have a doc which is out of our control, so it may change to something which would not work anymore. So I'd prefer to fix the link to a concrete version.

pkg/operator/credentials_test.go
```diff
@@ -0,0 +1,321 @@
...
+				},
+				Data: map[string][]byte{
+					"credentials": []byte("oksts"),
+				},
+			},
+			expectedCredReqName: types.NamespacedName{
+				Namespace: "openshift-cloud-credential-operator",
+				Name:      "aws-load-balancer-operator",
+			},
+			compareCredReq: func(credReq *cco.CredentialsRequest, providerSpec *cco.AWSProviderSpec) error {
```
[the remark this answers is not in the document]
alebedev87: Indeed why this way? It's so clever that even I forgot why I did so :) That's a good sign that this function deserves a comment.

I think my idea was based on the fact that I don't like to test things which don't change. We create a desired `CredentialsRequest` in `credentials.go`, there is only one copy of it so we either made a mistake in it or we don't. Repeating the same `CredentialsRequest` definition in `credentials_test.go` won't test anything - if the error is there we'll just repeat it in tests. The desired `CredentialsRequest` should be  (and it is) tested in the e2e test. So here I just tested the moving part of the desired `CredentialsRequest`. This approach helps to avoid the unnecessary repetition of the desired definition in each test case. But I agree it looks less intuitive.

README.md
```diff
@@ -53,26 +53,11 @@ podman push $IMG
...
-   ```
-5. Deploy the operator
+4. Deploy the operator
     ```bash
     make deploy
     ```
-6. The previous step deploys the conversion webhook, which requires TLS verification on the webhook client side. The
+5. The previous step deploys the conversion webhook, which requires TLS verification on the webhook client side. The
    manifests deployed through the `make deploy` command do not contain a valid certificate and key. You must provision a valid certificate and key through other tools.     
    If you run on OpenShift, you can use a convenience script, `hack/add-serving-cert.sh`, to enable [the service serving certificate feature](https://docs.openshift.com/container-platform/4.12/security/certificates/service-serving-certificate.html).
```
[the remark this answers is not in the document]
alebedev87: Same answer as for https://github.com/openshift/aws-load-balancer-operator/pull/109/files#r1337623230.
However this one can be updated to 4.13 as the docs don't seem to change since 4.12.

pkg/operator/credentials_test.go
```diff
@@ -0,0 +1,321 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/openshift/aws-load-balancer-operator/pkg/utils/test"
+)
+
+func TestProvisionCredentials(t *testing.T) {
```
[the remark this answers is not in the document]
alebedev87: Done.

alebedev87:
Second run over the comment, suggestions applied.

pkg/operator/credentials.go
alebedev87: > Not saying it needs to compare, but do you think users would benefit from a comparison? Just curious if you considered these scenarios and if you are okay with the implications.

I thought of this scenario. In a slightly different context but with the same result - what if the role ARN needs to be updated? I implemented it the way CCO people proposed in their [How to STS document](https://docs.google.com/document/d/1iFNpyycby_rOY1wUew-yl3uPWlE00krTgr9XHDZOTNo) because it's simple and agreed by the CCO people and reviewers of the document. However I'm still not 100% convinced because it's easy to get it wrong and end up in a situation when you have to do manual repairs.
Another point is the complexity of the comparison approach. Apart from the comparison itself which is a little more complex but still fine (as we have a lot of similar examples), there is the aspect of the secret's contents. If `CredentialsRequest` is updated, ALBO needs to make sure it reads the new contents not the old ones from the previous `CredentialsRequest` version.

test/e2e/operator_test.go
alebedev87: Yes, I think you are ahead of me here. Because I want this PR to be merged then I will need to update `AWSLoadBalancerController` API to add a new field for roleARN (another PR) and only after then I'll try all end to end.

> but it looks like it doesn't provision (create the E2E secret).

This may be related to what we discussed in Slack. If you created a ROSA STS cluster, it spawned 4.13.x OCP which doesn't have all `CredentialsRequest` fields this PR needs: `Spec.CloudTokenPath` is added in 4.14.0.

> Is STS supported for E2E testing yet? I think we need to provide the IAM Role ARN right?

Yes, we will have to update the CI to pass `ROLEARN` variable to the operator subscription. However the ROSA STS test case won't be able to use the new bootstrap mechanism because of the reason described above. So I suppose we will have to stay with the current CI for ROSA STS, that is, provision the secret in the CI.

alebedev87:
An AWS error while deleting TagretGroup, not enough log details to understand what happened:
```
{"level":"info","ts":1695760015.1396067,"logger":"controllers.ingress","msg":"deleting targetGroup","arn":"arn:aws:elasticloadbalancing:us-east-2:460538899914:targetgroup/k8s-awsloadb-echoserv-e6168287e8/e5c7e04303a2a230"}
{"level":"error","ts":1695760035.373432,"logger":"controller-runtime.manager.controller.ingress","msg":"Reconciler error","name":"multi-ingress","namespace":"","error":"failed to delete targetGroup: timed out waiting for the condition"}
{"level":"info","ts":1695760035.3787587,"logger":"controllers.ingress","msg":"successfully built model","model":"{\"id\":\"multi-ingress\",\"resources\":{}}"}
{"level":"info","ts":1695760036.006575,"logger":"controllers.ingress","msg":"deleting targetGroup","arn":"arn:aws:elasticloadbalancing:us-east-2:460538899914:targetgroup/k8s-awsloadb-echoserv-e6168287e8/e5c7e04303a2a230"}
{"level":"error","ts":1695760056.2372942,"logger":"controller-runtime.manager.controller.ingress","msg":"Reconciler error","name":"multi-ingress","namespace":"","error":"failed to delete targetGroup: timed out waiting for the condition"}
{"level":"info","ts":1695760056.2475882,"logger":"controllers.ingress","msg":"successfully built model","model":"{\"id\":\"multi-ingress\",\"resources\":{}}"}
{"level":"info","ts":1695760056.8230236,"logger":"controllers.ingress","msg":"deleting targetGroup","arn":"arn:aws:elasticloadbalancing:us-east-2:460538899914:targetgroup/k8s-awsloadb-echoserv-e6168287e8/e5c7e04303a2a230"}
{"level":"error","ts":1695760077.0477169,"logger":"controller-runtime.manager.controller.ingress","msg":"Reconciler error","name":"multi-ingress","namespace":"","error":"failed to delete targetGroup: timed out waiting for the condition"}
```

/test e2e-aws-operator

pkg/operator/credentials.go
alebedev87: Slack discussion: https://redhat-internal.slack.com/archives/C04KRSVAZ37/p1695799843775839.

test/e2e/operator_test.go
gcs278: Ack, I'm good with addressing this later.

> This may be related to what we discussed in Slack. If you created a ROSA STS cluster, it spawned 4.13.x OCP which doesn't have all CredentialsRequest fields this PR needs: Spec.CloudTokenPath is added in 4.14.0.

It was a 4.14.0-rc.1 STS cluster. I was just pointing out the E2E test doesn't provide the IAM Role ARN, so I think CCO can't create the E2E test secret. Sounds like we will address this later.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -31,59 +31,9 @@ spec:
...
-      ```bash
-      oc get secret -n aws-load-balancer-operator aws-load-balancer-operator
-      ```
-
-    After this the operator can be installated through the console or through the command line.
-
-    #### Instructions for STS cluster (inclusing ROSA in STS mode)
+    #### Instructions for STS cluster (including ROSA in STS mode)
 
     [Bootstrapping AWS Load Balancer Operator on Security Token Service cluster](https://docs.openshift.com/container-platform/latest/networking/aws_load_balancer_operator/installing-albo-sts-cluster.html#nw-bootstra-albo-on-sts-cluster_albo-sts-cluster).
```
gcs278, line 39: *unrelated nit*

Want to fix [line 42](https://github.com/openshift/aws-load-balancer-operator/pull/109/files#diff-dd5456da7fbaba7c98ed1f25f52b43479bfc015133b17ebae0972c1ad2e4eb4bR42): `VPC and subnets` to be `VPC and Subnets` to match the capitalization of the heading?
```
Additional documentation can be found in [VPC and subnets](https://github.com/openshift/aws-load-balancer-operator/blob/main/docs/prerequisites.md#vpc-and-subnets).
```

gcs278 (commented):
Looks good. A couple more minor comments.

pkg/operator/credentials.go
alebedev87: Yes, they seem to like the approach and will update the reference doc accordingly. Let me change the part then to add a comparison+update of credentialsrequest.

docs/prerequisites.md
```diff
@@ -1,93 +1,88 @@
...
 
-4. Apply the secrets to your cluster:
+    **Note**: Write down the role ARN for the aws-load-balancer-operator in the output, it will be needed for the next step.
 
-    ```bash
-    ls manifests/*-credentials.yaml | xargs -I{} oc apply -f {}
-    ```
-
-5. Verify that the operator's credentials secret is created:
+3. Install the operator using the OpenShift OperatorHub WebUI or by running the following commands:
```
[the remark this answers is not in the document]
alebedev87: Something with my rebase as I remember fixing that one. Let me do it again.

docs/prerequisites.md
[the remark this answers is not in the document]
alebedev87: Interesting, I think we need to clarify this with the OLM guys and if the default channel is indeed used as default :) we can remove it from the docs.

What I thought the default channel was doing is just setting the checkbox in OperatorHub UI:
image



[The OLM documentation](https://olm.operatorframework.io/docs/tasks/creating-operator-bundle/#contents-of-annotationsyaml-and-the-dockerfile) about the default channel annotation:
> The label operators.operatorframework.io.bundle.channel.default.v1 reflects the default channel an operator should be subscribed to when installed from a registry. This label is optional if the default channel has been set by previous bundles and the default channel is unchanged for this bundle.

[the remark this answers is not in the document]
alebedev87: Slack thread: https://redhat-internal.slack.com/archives/C01GJKL46A2/p1695890348693129.

test/e2e/operator_test.go
alebedev87: Yes, I think to attack the e2e once the CI will be ready (adding `ROLEARN` envvar), so followup PRs.

pkg/operator/credentials.go
alebedev87: Done in https://github.com/openshift/aws-load-balancer-operator/pull/109/commits/f1a89d8c1a3bd2b1d39eaad8d9f06e8b6866ca1a.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: Done in https://github.com/openshift/aws-load-balancer-operator/commit/f1a89d8c1a3bd2b1d39eaad8d9f06e8b6866ca1a.

pkg/utils/resource/update/update.go
gcs278, line 1: This is subjective, but, to me, it seems a little odd to place functions for `CredentialsRequest` outside of this package, based on other NE repos layouts. I see the readability of a package called `update`, but if we are trying to converge on styles, the ingress operator generally seems like it has a Go file representing object which houses an `ensure<Object>`, `create<Object>`, `current<Object>`, `desired<Object>`, `update<Object>`, and/or `<object>Changed`.

Here's some examples:
- https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/canary/daemonset.go
- https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/canary/namespace.go
- https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/gatewayapi/crds.go
- https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/gatewayclass/subscription.go
- https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/ingress/cluster-role.go

I don't think this has to be an exact science, but you could have a file `credentials_request.go` (or leave it in `credentials.go`) that would represent the operations for the CR object.

Am I missing a benefit of this pattern? And what are your thoughts on trying to converge on style? Agree or disagree?

pkg/utils/resource/update/update.go
```diff
@@ -0,0 +1,72 @@
...
+
+	cco "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/aws-load-balancer-operator/pkg/utils"
+)
+
+// CredentialsRequest updates the given CredentialsRequest to the desired state if needed.
+// Returns true if the CredentialsRequest was updated, false otherwise.
+func CredentialsRequest(ctx context.Context, client client.Client, current *cco.CredentialsRequest, desired *cco.CredentialsRequest) (bool, error) {
```
gcs278, line 15: Along with my style comment above, more specifically I feel like this could be better classified as an `ensure` operation vs. an `update`. Ensure implying it'll ensure it will get the object into the right state, but it may not need to update it, whereas update implies its going to always update it (which sometimes it doesn't here).

pkg/utils/resource/update/update.go
```diff
@@ -0,0 +1,72 @@
...
+	if changed, err := IsCredentialsRequestChanged(current, desired); err != nil {
+		return false, err
+	} else if !changed {
+		return false, nil
+	}
+
+	updated := current.DeepCopy()
+	updated.Name = desired.Name
```
gcs278, line 23: :exploding_head: You can update the name and namespace of an object after it's been created? I didn't think it was possible, I thought they were immutable.
Also, I don't see a unit test for either namespace or name change, I think that would be important to test.

pkg/utils/resource/update/update.go
```diff
@@ -0,0 +1,72 @@
...
+	if err := client.Update(ctx, updated); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// IsCredentialsRequestChanged compares the current and desired states of a CredentialsRequest.
+// Returns true of current start is different from the desired, false otherwise.
+func IsCredentialsRequestChanged(current, desired *cco.CredentialsRequest) (bool, error) {
+	if current.Name != desired.Name {
```
gcs278, line 35: Like I mentioned above, I don't think name and namespace is ever possible to be different. I am missing something?

pkg/utils/resource/update/update.go
```diff
@@ -0,0 +1,72 @@
...
+		return true, nil
+	}
+
+	if current.Namespace != desired.Namespace {
+		return true, nil
+	}
+
```
gcs278, line 42: Did you consider https://pkg.go.dev/github.com/google/go-cmp/cmp? I think it works nicely for simple comparisons, but not so sure about more advanced. 

Some examples:
- https://github.com/openshift/cluster-ingress-operator/blob/6cd3b6709eff0026ad814921b40c4b02c32fb38c/pkg/operator/controller/ingress/internal_service.go#L116
- https://github.com/openshift/cluster-ingress-operator/blob/d7ffd6e7a9cd89805e273b37fff6e6cca6298677/pkg/operator/controller/gatewayapi/crds.go#L115

pkg/utils/resource/update/update.go
```diff
@@ -0,0 +1,72 @@
...
+	if !utils.EqualStrings(current.Spec.ServiceAccountNames, desired.Spec.ServiceAccountNames) {
+		return true, nil
+	}
+
+	if current.Spec.CloudTokenPath != desired.Spec.CloudTokenPath {
+		return true, nil
+	}
+
+	currentAwsSpec := cco.AWSProviderSpec{}
+	err := cco.Codec.DecodeProviderSpec(current.Spec.ProviderSpec, &currentAwsSpec)
```
gcs278, line 56: Just curious, do you need to decode provider spec? What happens when you compare them as Raw?

gcs278 (commented):
Just a couple of minor comments and questions about style.

pkg/utils/resource/update/update.go
alebedev87: I agree, it makes sense to rename it to `credentials_request.go`.

pkg/utils/resource/update/update.go
alebedev87: I didn't want to change the logic as I reintegrated it back to the existing code, so I didn't want any change in the behavior. I think it was considered at the time of the creation. I'm fine with this more verbose approach as it explicitly states what we are interested in and let the rest be defaulted or ignored.

Btw, I'm always a little surprised how wide spread it is in the production code taking into account their note in the package description:
> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values.

pkg/utils/resource/update/update.go
alebedev87: I thought of the readability, the decoded diff is more human readable. However I tried raw too, it has a small glitch which I didn't dig for the sake of time as I was fine with the decoded approach - there is a new line which gets inserted somewhere in the CCO code.

pkg/utils/resource/update/update.go
alebedev87: Looking at the ensuring functions in this repo ([example](https://github.com/openshift/aws-load-balancer-operator/blob/main/pkg/controllers/awsloadbalancercontroller/credentials_request.go#L39)): it has to do `get-create/update-get`. At first I thought of moving the `ensureCredentialsRequest` function into this package but that would need too many adaptations to make it work in both: main and controller. So I took the function which needs the least changes for the existing code but does its job for the new code.

docs/prerequisites.md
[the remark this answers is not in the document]
alebedev87: Done.

pkg/utils/resource/update/update.go
alebedev87: Renamed to `credentials_request.go`.

pkg/utils/resource/update/update.go
alebedev87: Indeed, it's unclear what was expected ho happen here if either of them would be different. I removed these assignments and added some logic to treat the name/namespace mismatch as an input error.

alebedev87:
I noticed an early restart of the operator container. It's the liveness probe which is set to do 3 attempts with 20s interval. This is not enough as the timeout set for the secret waiting is 5 minutes. I set up a startupProbe to let the container fail if the secret was not provisioned.

pkg/utils/resource/update/update.go
gcs278: Okay - I was a little confused, I didn't realize you needed a public function between the `awsloadbalancercontroller` and `operator` package. Makes more sense now.

pkg/utils/resource/update/update.go
gcs278: Yea I see now. Seems reasonable. There's some refactoring you could probably do, but would require an overhaul. I think this is acceptable for now.

I do find it bit funky that the function name `CredentialsRequest` doesn't completely describe what the function does, but it needs the package name `update` to aptly describe it. If you are local, like in `credentials_request_test.go`, then you won't call the function with the package name (`update.CredentialsRequest` vs. `CredentialsRequest`). Is this pattern used in other places?

docs/prerequisites.md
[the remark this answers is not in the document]
gcs278: Seems like we are converging on just leaving it in. Sorry for the run around, but I appreciate the investigation.

pkg/utils/resource/update/update.go
gcs278: Okay I see, it's debatable. On the side of brevity, you are right, I agree with https://go.dev/doc/effective_go#package-names. But in the examples I see, the package name was the "noun" and the function was the "verb"; whereas, this is backwards (i.e. package `bufio` and function `reader` vs. package `update` and function `CredentialsRequest`).

I don't think this is important enough to hold up the PR. I'll LGTM and approve and you can proceed how you wish.

gcs278:
Thanks for all of the responses! I learnt a lot reviewing this PR. Feel free to continue to update, but as is, I feel this PR is ready to go.

/lgtm
/approve

alebedev87:
@gcs278 : squashed the commits which were the result of the review:
- waiting on the first AWS call (AWS credentials turned to be not usable immediately)
- update of `CredentialsRequest` instead of just create
- startupProbe

docs/prerequisites.md
```diff
@@ -1,93 +1,88 @@
...
+    apiVersion: operators.coreos.com/v1
+    kind: OperatorGroup
+    metadata:
+      name: aws-load-balancer-operator
+      namespace: aws-load-balancer-operator
+    spec:
+      targetNamespaces: []
+    EOF
+
+    $ ROLEARN="<iam-role-arn>"
```
lihongan, line 65: Can we provide more info or detailed commands for getting the `ROLEARN` ? Is this only required for STS cluster ?

docs/prerequisites.md
```diff
@@ -132,28 +127,42 @@ If you prefer not to use `ccoctl`, or your system doesn't support it, the AWS CL
...
+      namespace: aws-load-balancer-operator
+    spec:
+      channel: stable-v1
+      name: aws-load-balancer-operator
+      source: redhat-operators
+      sourceNamespace: openshift-marketplace
+      config:
+        env:
+        - name: ROLEARN
+          value: "${OPERATOR_ROLE_ARN}"
```
lihongan, line 163: maybe should update the value as 
```
value: "${ROLEARN}"
```
And is this ENV required on STS cluster only ?

alebedev87: Updated to `$ROLEARN` to be consistent with `ccoctl` option above.

docs/prerequisites.md
alebedev87: Added a step above to set `ROLEARN` environment variable.

docs/prerequisites.md
```diff
@@ -1,93 +1,91 @@
...
-    of secrets in a **manifests** directory that is required
-    by the **aws-load-balancer-operator**.
+    of secrets in a `manifests` directory, which are not needed by the operator.
 
-4. Apply the secrets to your cluster:
+3. Write down the role ARN for the aws-load-balancer-operator from the output of `ccoctl`:
 
     ```bash
-    ls manifests/*-credentials.yaml | xargs -I{} oc apply -f {}
+    $ ROLEARN=$(\grep -Po 'arn:aws:iam[0-9a-z/:\-_]+' "${CCOCTL_OUPUT}")
```
gcs278, line 52:
```suggestion
    $ ROLEARN=$(grep -Po 'arn:aws:iam[0-9a-z/:\-_]+' "${CCOCTL_OUTPUT}")
```

I also have a slight issue with this current automation. Per my other comments, I placed `--credentials-requests-dir=<path-to-credrequests-dir>` into my existing credentials request dir created by `ccoctl` during installation.

This works fine, but re-applies the credentials and the output contains all of the RoleARNs, which would break this as it would grab all of the IAM roleARNs.

Maybe it was a noob move on my part? Or should we provide a hard-coded `<path-to-credrequests-dir>` to imply it's new and not it's supposed to be an existing directory?

docs/prerequisites.md
```diff
@@ -1,93 +1,91 @@
...
-    ```bash
-    curl --create-dirs -o <path-to-credrequests-dir>/cr.yaml https://raw.githubusercontent.com/openshift/aws-load-balancer-operator/main/hack/operator-credentials-request.yaml
-    ccoctl aws create-iam-roles \
+   $ curl --create-dirs -o <path-to-credrequests-dir>/cr.yaml https://raw.githubusercontent.com/openshift/aws-load-balancer-operator/main/hack/operator-credentials-request.yaml
+   $ CCOCTL_OUPUT=$(mktemp)
+   $ ccoctl aws create-iam-roles \
         --name <name> --region=<aws_region> \
         --credentials-requests-dir=<path-to-credrequests-dir> \
-        --identity-provider-arn <oidc-arn>
+        --identity-provider-arn <oidc-arn> 2>&1 | tee "${CCOCTL_OUPUT}"
```
gcs278, line 37:
```suggestion
        --identity-provider-arn <oidc-arn> 2>&1 | tee "${CCOCTL_OUTPUT}"
```

docs/prerequisites.md
alebedev87: `\grep` was to prevent the usage of the bash aliases, for instance, I use a grep alias which may give a different output. But for sake brevity, I agree it's not necessary.

> Or should we provide a hard-coded <path-to-credrequests-dir> to imply it's new and not it's supposed to be an existing directory?

I see. Well, yes I think there are a lot of ways of how the end user may screw up these commands. I updated the step to:
```
3. Extract and verify the operator's role ARN from the output of `ccoctl` command:                                       

    ```bash
    $ ROLEARN=$(grep -Po 'arn:aws:iam[0-9a-z/:\-_]+' "${CCOCTL_OUTPUT}")
    $ echo "${ROLEARN}"
    arn:aws:iam::777777777777:role/<name>-aws-load-balancer-operator-aws-load-balancer-operator
    ```
```
in order to delegate some responsibilities to the end user :)

docs/prerequisites.md
alebedev87: Done.

docs/prerequisites.md
gcs278: Oh wow, TIL about avoiding bash aliases like that.

Okay, that works for me, thanks :+1:

PR #113, opened 2023-09-27
NE-1381: API - add STSIAMRoleARN field

alebedev87:
**Context**
According to [Tokenized Cloud Auth Enablement For Operators EP](https://github.com/openshift/enhancements/pull/1339) CCO can now provision secrets on STS clusters too. However this EP is meant for the operators only, the operands are supposed to be managed via their APIs.

**Goal**
The goal of this PR is to add a new field for the Role ARN which is supposed to be added to `CredentialsRequest` CR created by the operator for its operand (AWS LoadBalancer Controller).

PR #97, opened 2023-08-02
OCPBUGS-15609: docs - add AWS CLI commands for STS instructions

alebedev87:
This PR updates the documentation about the manual configuration for STS clusters. The main goal is to provide AWS CLI commands as alternative to `ccoctl` commands because the latter
- is not available on some OSes (Mac, Windows)
- cannot be installed in a convenient way
- not transparent about the AWS permissions it takes from the user's AWS credentials

Makefile's `generate` target is updated to generate the missing permission policy for the operator. This permission policy is required by AWS to create a IAM role.

Manual test on ROSA STS cluster:
```
$ oc get awsloadbalancercontroller cluster -o yaml | yq .spec
credentials:
  name: aws-load-balancer-controller-cluster
ingressClass: alb
subnetTagging: Auto

$ oc -n aws-load-balancer-operator get pods
NAME                                                            READY   STATUS    RESTARTS   AGE
aws-load-balancer-controller-cluster-64697c78bf-chdkn           1/1     Running   0          16m
aws-load-balancer-operator-controller-manager-c67f6b5cf-gvxgs   2/2     Running   0          166m

$ oc -n aws-load-balancer-operator get pods aws-load-balancer-controller-cluster-64697c78bf-chdkn -o yaml | yq .spec.volumes
- name: aws-credentials
  secret:
    defaultMode: 420
    secretName: aws-load-balancer-controller-cluster
    
$ oc -n aws-load-balancer-operator get pods aws-load-balancer-operator-controller-manager-c67f6b5cf-gvxgs -o yaml | yq .spec.volumes
- name: aws-credentials
  secret:
    defaultMode: 420
    items:
      - key: credentials
        path: credentials
    secretName: aws-load-balancer-operator
  
$ oc -n aws-load-balancer-operator get secret aws-load-balancer-operator -o yaml | yq .data.credentials | base64 -d 
[default]
sts_regional_endpoints = regional
role_arn = arn:aws:iam::269733383066:role/albo-operator
web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token

$ oc -n aws-load-balancer-operator get secret aws-load-balancer-controller-cluster -o yaml | yq .data.credentials | base64 -d 
[default]
sts_regional_endpoints = regional
role_arn = arn:aws:iam::269733383066:role/albo-controller
web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
```

PR #114, opened 2023-10-17
OCPBUGS-19757: bump controller iam policy to v2.4.7

alebedev87:
**Context**
AWS changed the required permissions to create a LB target group: [upstream issue describing the change](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2692).

**Why didn't we catch this in CI?**
The problem started to be visible after [AWS CLI commands were published](https://github.com/openshift/aws-load-balancer-operator/pull/97). With these commands the upstream IAM policy started to be used instead of [the generated controller's credentials request](https://github.com/openshift/aws-load-balancer-operator/blob/main/hack/controller/controller-credentials-request.yaml). The difference between the two is that the latter is minified to meet [the AWS limit for inline policies](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length). The minified CredentialsRequest has a broader set of permissions and since it's [used in the CI to create the controller's role](https://github.com/openshift/release/blob/3777337d4c56834709f45de404911575444f744e/ci-operator/step-registry/aws-load-balancer/pre-install-rosa/aws-load-balancer-pre-install-rosa-commands.sh#L53-L54) we never saw the missing permissions.

**Changes**
The main change is a bump of the upstream IAM policy (`assets` directory). However to catch similar problems in the future we need to have the controller's CredentialsRequest to be the same as the upstream's IAM policy. For this we need to stop minifying. This in turn requires significant changes in `iamctl` binary to be able to support:
-  multiple resources from AWS statements because `CredentialsRequest`'s resource field is not a slice (`-s` flag added)
- conditions from AWS statements
- customization of the IAM Policy go function (`-f` flag)

**Result**
The IAM policy is bumped enabling the users of AWS CLI command to use the right set of permissions for the controller's role. The controller's `CredentialsRequest` is not minified anymore so that the role created in ROSA e2e using `ccoctl` is as similar as possible to the upstream.

**Note**
[The IAM policy generated for non-STS clusters](https://github.com/openshift/aws-load-balancer-operator/blob/main/pkg/controllers/awsloadbalancercontroller/iam_policy.go#L10) must still be minified as the AWS limit is still there. A new IAM policy for STS will be added in this PR to be used in case [the STS role is given to the controller](https://github.com/openshift/aws-load-balancer-operator/pull/113).


Makefile
```diff
@@ -127,13 +131,27 @@ vet: ## Run go vet against code.
 
 .PHONY: iamctl-gen
 iamctl-gen: iamctl-build iam-gen
-	$(IAMCTL_BINARY) -i $(IAMCTL_ASSETS_DIR)/iam-policy.json -o $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE) -p $(IAMCTL_GO_PACKAGE) -c $(IAMCTL_OUTPUT_CR_FILE)
-	go fmt -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
-	go vet -mod=vendor $(IAMCTL_OUTPUT_DIR)/$(IAMCTL_OUTPUT_FILE)
+	# generate controller's IAM policy without minify.
+	@# This policy is for STS clusters as it's turned into a role policy which is limited to 10240 by AWS.
```
gcs278, line 135: Okay, so if I understand this correctly, you *could* minify for STS clusters as well, but by not minifying, you help expose issues with the non-minified IAM Policy, of which QE uses to install clusters and is documented `Option 2. Using the AWS CLI` in `install.md`?

gcs278:
I just had a clarifying question, but thanks for the very detailed description. It took me couple of re-reads, but I think I understand your motiviation.
/approve

Makefile
alebedev87: Yes, exactly.

PR #116, opened 2023-10-27
OCPBUGS-22344: address CVE-2023-44487

alebedev87:
What was done to address CVE-2023-44487:
- bump controller-runtime v0.15.3
  - bump of `x/net` and kube api were done along the way
- bump the operand image to the latest version
- disable HTTP2 for kube-rbac-proxy container
- disable HTTP2 for webhook server


bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -249,7 +249,8 @@ spec:
                 - --upstream=http://127.0.0.1:8080/
                 - --logtostderr=true
                 - --v=0
-                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
+                - --http2-disable
+                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
```
gcs278, line 253: I noticed external-dns-operator [uses](https://github.com/openshift/external-dns-operator/blob/b7633e786424c8209b4fe1fccb011b212ea48357/bundle/manifests/external-dns-operator.clusterserviceversion.yaml#L484) `quay.io/openshift/origin-kube-rbac-proxy:latest`.

Should we be using quay.io images instead of other external images? Also, `latest` tag vs. pinning to specific version is different.

main.go
```diff
@@ -99,6 +102,7 @@ func main() {
 	flag.StringVar(&namespace, "namespace", "aws-load-balancer-operator", "The namespace where operands should be installed")
 	flag.StringVar(&image, "image", "quay.io/aws-load-balancer-operator/aws-load-balancer-controller:latest", "The image to be used for the operand")
 	flag.StringVar(&trustedCAConfigMapName, "trusted-ca-configmap", "", "The name of the config map containing TLS CA(s) which should be trusted by the controller's containers. PEM encoded file under \"ca-bundle.crt\" key is expected.")
+	flag.BoolVar(&webhookDisableHTTP2, "webhook-disable-http2", false, "Disable HTTP2 for the webhook server.")
```
gcs278, line 105: https://github.com/openshift/external-dns-operator/pull/204#discussion_r1388577353

```suggestion
	flag.BoolVar(&webhookDisableHTTP2, "webhook-disable-http2", false, "Disable HTTP/2 for the webhook server.")
```

gcs278 (commented):
Questions and a nit, but otherwise looks good.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -274,11 +275,12 @@ spec:
                 - --image=$(RELATED_IMAGE_CONTROLLER)
                 - --namespace=$(TARGET_NAMESPACE)
                 - --trusted-ca-configmap=$(TRUSTED_CA_CONFIGMAP_NAME)
+                - --webhook-disable-http2
                 command:
                 - /manager
                 env:
                 - name: RELATED_IMAGE_CONTROLLER
-                  value: quay.io/aws-load-balancer-operator/aws-load-balancer-controller@sha256:f359f63169432cf8029b189f6605f21d660f0f805b91d55fe964185d0b8ae0b2
+                  value: quay.io/aws-load-balancer-operator/aws-load-balancer-controller@sha256:d8b5e9a91aca2a4a4de7f9bd2b614c5ba3d4cc62fa8967e94e9539fd7c1940a9
```
Miciah, line 283: For reference, here's the corresponding manifest: https://quay.io/repository/aws-load-balancer-operator/aws-load-balancer-controller/manifest/sha256:d8b5e9a91aca2a4a4de7f9bd2b614c5ba3d4cc62fa8967e94e9539fd7c1940a9

The manifest labels reference the expected aws-load-balancer-controller commit: `"vcs-ref"="0ae6e1a96d44f67fe852e91ec31a79d138022cd2" "vcs-type"="git" "vcs-url"="https://github.com/openshift/aws-load-balancer-controller"`, which is <https://github.com/openshift/aws-load-balancer-controller/commit/0ae6e1a96d44f67fe852e91ec31a79d138022cd2>.

main.go
alebedev87: Done.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: Comments added to the manager manifest.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
Miciah: Will we also need a substitution like we have for external-dns-operator? https://github.com/openshift/release/blob/f4f1bf0fe4f6af1eb6525431de841db8ba9a22b8/ci-operator/config/openshift/external-dns-operator/openshift-external-dns-operator-release-4.16.yaml#L36-L37

alebedev87: Hm, I'm not sure what this substitution gives us. I see only the concrete OCP tag instead of `latest`. The base image is still RHEL (btw this is a good thing compared to gcr.io image as we would like to be closer to what we ship). I'm fine with the latest tag because it avoids a drift between the local testing and CI testing plus doesn't bloat the CI config.

Miciah: Hm, so external-dns-operator is using `quay.io/openshift/origin-kube-rbac-proxy:latest` despite the substitution that is specified in the CI configuration?  In that case, I'm not sure what the point of the substitution is.  I figured it would substitute in the kube-rbac-proxy image from the release payload.  Does such substitution only happen for branches other than the main branch?

alebedev87: > so external-dns-operator is using quay.io/openshift/origin-kube-rbac-proxy:latest despite the substitution that is specified in the CI configuration?

There are 3 places: developer's machine where what's committed in GitHub repository is used as the image, CI where a substitution may happen before the test is run, CPaaS where the registry has to change to the redhat.registry.com.
I think that keeping these 3 cases as close as possible to each other is a goal. However some diffs are inevitable as some code is open (for contributions) and some is closed (for Red Hat products).

> I figured it would substitute in the kube-rbac-proxy image from the release payload.

Yes, I also think it's the purpose of the substitution. However I greped on the CI config and only 3 operators do the substitution for `kube-rbac-proxy` and 2 of them done by me :) 

> Does such substitution only happen for branches other than the main branch?

Well, indeed, `main` doesn't have. I [removed it](https://github.com/openshift/release/commit/23de4d9d68180c559bbc400d78476ebd0e997b7d#diff-8f9a46e5fe5d27abff924e731c5d0578023e92a751e5187512510e8bd72369eb) as part of the works to reduce the dependency on all things OpenShift. 


I don't see a reason of why we should replace kube-rbac-proxy image with the payload one. I'll do a PR to remove it to simplify the CI config and to reduce the drift between the local development and CI testing.

alebedev87: Asked to opt out from the ART automation again: https://redhat-internal.slack.com/archives/CB95J6R4N/p1701723839704069.

Miciah:
Thanks!
/approve
/lgtm

PR #117, opened 2023-11-02
NO-JIRA: Use trust policy term

alebedev87:
Replace `trusted policy` term with the correct [`trust policy`](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_terms-and-concepts.html#term_trust-policy). Follow-up of [a PR in the docs](https://github.com/openshift/openshift-docs/pull/67219#discussion_r1379837558).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -130,12 +130,12 @@ Its contents are identical to the ones requested by **aws-load-balancer-operator
 
 #### Option 2. Using the AWS CLI
 
-1. Generate a trusted policy file using your identity provider (e.g. OpenID Connect):
+1. Generate a trust policy file using your identity provider (e.g. OpenID Connect):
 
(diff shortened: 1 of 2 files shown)


alebedev87:
/label docs-approved
/label px-approved
/label qe-approved

Minor docs change, no need for a round of approvals.

candita:
@alebedev87 please keep an eye on this to see if it merges.  The NO-JIRA should have added a jira/valid-reference but it doesn't, even if I comment /jira refresh.

candita:
A test, to see if we can do this manually:
/jira valid-bug

candita:
@alebedev87 do you want to put the labels you need here?  It is approved and lgtm, it's just that the NO-JIRA doesn't work as intended.

PR #122, opened 2023-11-15
OCPBUGS-36641: Add versioning and branching documentation

alebedev87:
This PR:
- Documents the versioning model, repository branch organization, OLM channels, OpenShift compatibility, and support model.
- Aims to guide users on selecting the appropriate ALBO version for their OpenShift environment.

### DIFF
--- a/docs/versioning.md
+++ b/docs/versioning.md
@@ -1,0 +1,48 @@
+# Versioning and Branching in AWS Load Balancer Operator
+
+The AWS Load Balancer Operator follows the semantic versioning, for any given release `X.Y.Z`:
+* an X (major) release indicates a set of backwards-compatible changes. Changing X means there's a breaking change.
+* a Y (minor) release indicates a minimum feature set. Changing Y means the addition of a backwards-compatible feature.
+* a Z (patch) release indicates minimum set of bugfixes. Changing Z means a backwards-compatible change that doesn't add functionality.
+
+## Branches
+
+The AWS Load Balancer Operator repository contains two types of branches: the `main` branch and `release-X.Y` branches.
+
+The main branch is where development happens. All the latest code, including breaking changes, happens on `main`.
+The `release-X.Y` branches contain stable, backwards compatible code. Every minor (`X.Y`) release, a new such branch is created.
+
+## Channels
+
+The AWS Load Balancer Operator's releases get published in two types of [OLM channels](https://olm.operatorframework.io/docs/glossary/#channel): the minor `release-vX.Y` and the major `release-vX`.
+
+The minor channels contain patch releases. The major channels contain all patch releases from all minor channels.
+
+## OpenShift Version Compatibility
+
+The table below illustrates the OpenShift versions for which various AWS Load Balancer Operator releases were intended:
+
+| OCP version | AWS Load Balancer Operator branch  | AWS Load Balancer Operator OLM channel |
+| :---------: | :-------------------------------:  | :------------------------------------: |
+| 4.16        | release-1.1                        | stable-v1.1, stable-v1                 |
+| 4.15        | release-1.1                        | stable-v1.1, stable-v1                 |
+| 4.14        | release-1.1                        | stable-v1.1, stable-v1                 |
+| 4.13        | release-1.0                        | stable-v1.0, stable-v1                 |
+| 4.12        | release-0.2                        | stable-v0.2, stable-v0                 |
+| 4.11        | release-0.1                        | stable-v0.1, stable-v0                 |
+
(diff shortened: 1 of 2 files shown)


docs/versioning.md
```diff
@@ -0,0 +1,35 @@
...
+The AWS Load Balancer Operator repository contains two types of branches: the `main` branch and `release-X.Y` branches.
+
+The main branch is where development happens. All the latest code, including breaking changes, happens on `main`.
+The `release-X.Y` branches contain stable, backwards compatible code. Every minor (`X.Y`) release, a new such branch is created.
+
+## Channels
+
+The AWS Load Balancer Operator's releases get published in two types of [OLM channels](https://olm.operatorframework.io/docs/glossary/#channel): the minor `release-vX.Y` and the major `release-vX`.
+
+The minor channels contain patch releases. The major channels contain all patch releases from all minor channels.
```
lihongan, line 19: I'm doubt if we should keep the major channels? I cannot see its use case, it that for upgrade automatically?

alebedev87: > is that for upgrade automatically?

Yes. The idea was to simplify the upgrade and to minimize the configuration on the client side. Since we often publish new minor channels (practically every new feature) the client has to change its subscription to explicitly mention new channel. Also, the OCP docs don't have to updated every time we publish a new feature.

lihongan:
cc @melvinjoseph86 @rhamini3

ShudiLi:
/lgtm
thanks

docs/versioning.md
```diff
@@ -0,0 +1,37 @@
...
+| 4.15        | release-1.1                        | stable-v1.1, stable-v1                 |
+| 4.14        | release-1.1                        | stable-v1.1, stable-v1                 |
+| 4.13        | release-1.0                        | stable-v1.0, stable-v1                 |
+| 4.12        | release-0.2                        | stable-v0.2, stable-v0                 |
+| 4.11        | release-0.1                        | stable-v0.1, stable-v0                 |
+
+## Support model
+
+* **Full support**: AWS Load Balancer Operator releases installed on the designated OpenShift version (see [OpenShift Version Compatibility](#openshift-version-compatibility)).
+* **Best effort**: AWS Load Balancer Operator release installations or upgrades outside of the designated OpenShift version (see [OpenShift Version Compatibility](#openshift-version-compatibility)).
```
gcs278, line 37: Is it worth defining "Full Support" and "Best Effort"? I mostly understand from the context of the words, but we don't really define them.

I assume Full Support = Fully compatible. All supported features expected to work.
I assume Best Effort = Not guaranteed to be fully compatible. Some features may not work.

I think Best Effort's a bit of a vague zone to be in and maybe even a paradox. I wonder if it really makes sense. It's kind of a "yea you can try it, it might work, but we don't really support it". Is that really our "Best Effort" lol?

Should we be more opinionated and redefine "Best Effort" in terms of "Support"? Like "Full Support" and "Not Supported" or "No Support". But then describe "No Support" as "It will likely install, but some features may not function correctly on this version". I can't say this is better, but maybe is more specific regarding support.

Just thinking out loud, I don't have any major problems with your choice of words.

alebedev87: Right, "Best Effort" sounds vague. I think I wanted to sound "more welcoming" knowing this practice exist among our users. I removed "Best Effort" completely and described what "Full Support" means. Also added a new table to show which versions are "End of Life".

docs/versioning.md
```diff
@@ -0,0 +1,48 @@
...
+
+The minor channels contain patch releases. The major channels contain all patch releases from all minor channels.
+
+## OpenShift Version Compatibility
+
+The table below illustrates the OpenShift versions for which various AWS Load Balancer Operator releases were intended:
+
+| OCP version | AWS Load Balancer Operator branch  | AWS Load Balancer Operator OLM channel |
+| :---------: | :-------------------------------:  | :------------------------------------: |
+| 4.16        | release-1.1                        | stable-v1.1, stable-v1                 |
```
gcs278, line 27: Can you add 4.17 while we are still working on this PR?

alebedev87: I will add it after the controller bump (`2.8.2`) along with a new olm channel. Like [I did in ExtDNSOperator](https://github.com/openshift/external-dns-operator/pull/233/files#diff-c08747565b40689f0a65a25b5d6e77bc49565ed7cef1a6d60e1f2c3d4ed4a8ff).

gcs278:
I think it's reasonable. We can change this in the future if we realize we need to adapt to a different support model, or language. 

/approve
/lgtm

PR #99, opened 2023-08-10
OCPBUGS-14846: update the controller image to use TLS1.2 as minumum version for webhook server

alebedev87:
This PR adds the aws-load-balancer-controller image containing the fix for the FIPS clusters whose maximum TLS version cannot be greater than 1.2.

The manual test can be found in the description of [the controller PR](https://github.com/openshift/aws-load-balancer-controller/pull/12).

PR #128, opened 2024-01-17
OCPBUGS-27023: add infrastructure annotations

alebedev87:
According to the Red Hat Operator's [Best practices](https://docs.engineering.redhat.com/display/CFC/Best_Practices) the infrastructure annotations are now required. The [old "infrastructure-features" annotation is deprecated](https://github.com/openshift/openshift-docs/pull/69579) hence I removed it.

### DIFF
--- a/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -66,9 +66,15 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    features.operators.openshift.io/disconnected: "false"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.1.0
     operatorframework.io/suggested-namespace: aws-load-balancer-operator
-    operators.openshift.io/infrastructure-features: '["proxy-aware"]'
     operators.operatorframework.io/builder: operator-sdk-v1.16.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/openshift/aws-load-balancer-operator
--- a/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
@@ -4,9 +4,15 @@ metadata:
   annotations:
     alm-examples: '[]'
     capabilities: Basic Install
+    features.operators.openshift.io/disconnected: "false"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.1.0
     operatorframework.io/suggested-namespace: aws-load-balancer-operator
-    operators.openshift.io/infrastructure-features: '["proxy-aware"]'
     repository: https://github.com/openshift/aws-load-balancer-operator
     support: Red Hat, Inc.
   name: aws-load-balancer-operator.v1.1.0


config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -4,9 +4,15 @@ metadata:
...
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: <1.1.0
     operatorframework.io/suggested-namespace: aws-load-balancer-operator
-    operators.openshift.io/infrastructure-features: '["proxy-aware"]'
```
Miciah, line 9: Can you describe in the commit message why you are adding the additional capabilities, and cite the relevant commits or Jira tickets related to the newly added capabilities?

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -66,9 +66,15 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
```
lihongan, line 70: The fips bug [OCPBUGS-14846](https://issues.redhat.com/browse/OCPBUGS-14846) has been fixed by PR https://github.com/openshift/aws-load-balancer-operator/pull/99 but that's for latest v1.1.0.
If this PR is back ported to old release then we need to update this annotation to "false" I think.

alebedev87: Got catch! Yes, I'll need to backport this PR down to 1.0. So the 1.0 PR needs to have `fips-compliant` set to `false`.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: The commit message changed.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
Miciah: I just want to double-check that this is correct.  

We set `CGO_ENABLED=0` in the Dockerfile: https://github.com/openshift/aws-load-balancer-operator/blob/1a2c320b27d4b5bb1a9c259db0796c462a11cc98/Dockerfile#L17

We don't set `CGO_ENABLED` in the Makefile: https://github.com/openshift/aws-load-balancer-operator/blob/1a2c320b27d4b5bb1a9c259db0796c462a11cc98/Makefile#L178

The Dockerfile is referenced here: https://github.com/openshift/release/blob/8703107e5b40d8559b2d7858350a88e3634762a6/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml#L17

Do we use the Dockerfile or the Makefile?  Is `CGO_ENABLED` set or not, and if it is, to what value?  

My understanding based on https://github.com/openshift/external-dns-operator/pull/213#discussion_r1458679687, is that we need `CGO_ENABLED=1` (or maybe not setting `CGO_ENABLED` at all is sufficient?) in order to use OpenSSL, and using OpenSSL (from UBI) is sufficient to claim FIPS compliance using the `features.operators.openshift.io/fips-compliant: "true"` annotation.  Do I understand correctly?

alebedev87: > Do we use the Dockerfile or the Makefile?

You found the right CI config, we use `Dockerfile` to build the operator image which is then [added to the OLM bundle](https://github.com/openshift/release/blob/8703107e5b40d8559b2d7858350a88e3634762a6/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml#L25-L27) which in turn is [deployed](https://github.com/openshift/release/blob/8703107e5b40d8559b2d7858350a88e3634762a6/ci-operator/config/openshift/aws-load-balancer-operator/openshift-aws-load-balancer-operator-main.yaml#L61) for each e2e test.

> Is CGO_ENABLED set or not, and if it is, to what value?

It's set to `0` which means that we use a statically built operator for the e2e tests. Also the users of the GitHub version of ALBO may use `Dockerfile` and build not FIPS compliant operator binary. Then what we deliver to the customers is built on CPaaS where [we don't set](https://gitlab.cee.redhat.com/cpaas-midstream/aws-load-balancer-operator/-/blob/aws-lb-optr-1.1-rhel-8/distgit/containers/aws-load-balancer-operator/Dockerfile.in?ref_type=heads#L10) `CGO_ENABLED` so the operator is built for the dynamic linkage.

> My understanding based on https://github.com/openshift/external-dns-operator/pull/213#discussion_r1458679687, is that we need CGO_ENABLED=1 (or maybe not setting CGO_ENABLED at all is sufficient?) in order to use OpenSSL, and using OpenSSL (from UBI) is sufficient to claim FIPS compliance using the features.operators.openshift.io/fips-compliant: "true" annotation. Do I understand correctly?

Default value for `CGO_ENABLED` is `1`, so it's enough to not set it. Whether the UBI image is fine the FIPS compliance or not is not 100% clear to me. I tend to say yes as it's RHEL based. Asked [a question](https://docs.google.com/presentation/d/1o3IowxHX6BsnxGkIInaQ0lBgnn_K5Ex8jxwCYCeNsqs/edit?disco=AAABGTECK14) in FIPS workshop slides.

Let me remove `CGO_ENABLED=0` from `Dockerfile`.

alebedev87: `CGO_ENABLED=0` removed from Dockerfile.

Miciah:
Hm, maybe removing `CGO_ENABLED=0` was not the right thing to do.  CI is now failing, with the aws-load-balancer-operator-controller-manager pod logging the following:
```
/manager: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /manager)
/manager: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /manager)
```
---
By the way, is there any reason not to set `terminationMessagePolicy: FallbackToLogsOnError` on all pod containers?  David has been making that change to a lot of other operators (see [OCPBUGS-28230](https://issues.redhat.com/browse/OCPBUGS-28230)) in order to make this sort of failure a little easier to spot, so I just now posted #131 to do the same for this operator.

Miciah:
Oh, do we need to resolve [OCPBUGS-24653 aws-load-balancer-operator fails FIPS check-payload scan](https://issues.redhat.com/browse/OCPBUGS-24653) before we can specify `features.operators.openshift.io/fips-compliant: "true"`?

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -66,9 +66,15 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    features.operators.openshift.io/disconnected: "false"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
```
candita, line 73: Only `proxy-aware` is set as a true infrastructure features in the previous version.  Can you explain why you mark `token-auth-aws` and `fips-compliant` as true here?

Miciah: The commit message now explains why these features are marked as true.

alebedev87:
`e2e-aws-rosa-operator` is failing due to [an ongoing issue](https://issues.redhat.com/browse/OSD-21663) with the OCP 4.16 candidate integration into ROSA.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: https://github.com/openshift/aws-load-balancer-operator/pull/133 fixed the FIPS compliance for ALBO. I think we have the right to set `true` now.

alebedev87:
A link to https://issues.redhat.com//browse/OCPBUGS-24653 has been added to the commit message.

Miciah:
Thanks!
/approve
/lgtm
/hold
for @candita.

bundle/manifests/aws-load-balancer-operator.clusterserviceversion.yaml
candita: What a phrase - "human obliviousness" !

candita:
LGTM
/unhold

PR #133, opened 2024-03-26
OCPBUGS-24653: Ensure FIPS compliance for operator image

alebedev87:
This commit addresses all the errors detected by [the check-payload tool](https://github.com/openshift/check-payload) by implementing the following changes:
- Replaced the base image with a non-UBI variant.
- Updated the builder image to be FIPS-capable.
- Enabled dynamic linkage for the operator binary (`CGO_ENABLED=1`).
- Added the `strictfipsruntime` tag to the operator binary.

Additionally, implemented `image-fips-scan` command for local FIPS compliance checks.

Result:
```
$ IMG=quay.io/alebedev/albo make image-fips-scan
sudo podman run --privileged registry.ci.openshift.org/ci/check-payload:latest scan operator --spec quay.io/alebedev/albo
I0326 08:50:55.098182       1 main.go:302] using embedded config
...
I0326 08:50:55.098589       1 main.go:102] "scan" version="0.3.1-91-ga0d1e0d3-dirty"
---- Successful run
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,8 @@ IAMCTL_OUTPUT_MINIFY_CR_FILE ?= ./hack/controller/controller-credentials-request
 # Built go binary path.
 IAMCTL_BINARY ?= ./bin/iamctl
 
+CHECK_PAYLOAD_IMG ?= registry.ci.openshift.org/ci/check-payload:latest
+
 
 .PHONY: all
 all: build
@@ -189,10 +191,14 @@ image-build: build test ## Build container image with the manager.
 image-push: ## Push container image with the manager.
 	${CONTAINER_ENGINE} push ${IMG}
 
+.PHONY: image-fips-scan
+image-fips-scan: image-build image-push
+	$(CONTAINER_ENGINE) run --privileged $(CHECK_PAYLOAD_IMG) scan operator --spec $(IMG)
+
 .PHONY: iamctl-build
 iamctl-build: fmt vet ## Build iamctl binary.
 	mkdir -p ./bin
-	cd ./cmd/iamctl && go build -mod=vendor -o $(IAMCTL_BINARY) . 
+	cd ./cmd/iamctl && go build -mod=vendor -o $(IAMCTL_BINARY) .
 	mv ./cmd/iamctl/$(IAMCTL_BINARY) ./bin/
 
 ##@ Deployment
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,5 @@
 # Build the manager binary
-FROM golang:1.19 as builder
-# TODO: Switch to UBI golang image when 1.19 is released
-# FROM registry.access.redhat.com/ubi8/go-toolset:latest
-
-WORKDIR /workspace
+FROM registry.access.redhat.com/ubi8/go-toolset:1.19 as builder
 
 # Copy the go source
 COPY main.go main.go
@@ -14,13 +10,11 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
+RUN GOOS=linux GOARCH=amd64 go build -tags strictfipsruntime -a -o /usr/bin/manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM registry.access.redhat.com/ubi8-minimal:latest
+FROM registry.redhat.io/rhel8-6-els/rhel:latest
 WORKDIR /
-COPY --from=builder /workspace/manager .
+COPY --from=builder /usr/bin/manager .
 
 USER 65532:65532
 


Miciah:
/assign
/approve
/lgtm
/hold
for @gcs278 to review.

Makefile
```diff
@@ -189,10 +191,14 @@ image-build: build test ## Build container image with the manager.
 image-push: ## Push container image with the manager.
 	${CONTAINER_ENGINE} push ${IMG}
 
+.PHONY: image-fips-scan
+image-fips-scan: image-build image-push
+	sudo $(CONTAINER_ENGINE) run --privileged $(CHECK_PAYLOAD_IMG) scan operator --spec $(IMG)
```
gcs278, line 196: Do we need the sudo? I noticed the other `$(CONTAINER_ENGINE)` commands don't have that, that seems a bit odd. Did you need that for your local machine?

Is this new command just for troubleshooting? Or will a downstream CI job be using this?

Makefile
```diff
@@ -189,10 +191,14 @@ image-build: build test ## Build container image with the manager.
 image-push: ## Push container image with the manager.
 	${CONTAINER_ENGINE} push ${IMG}
 
+.PHONY: image-fips-scan
+image-fips-scan: image-build image-push
```
gcs278, line 195: *nit* Do you need to do build and push in this step?

I had a bit of pain with this, because to build and push I use different registry auth files (one that can pull the images from registry.ci.openshift.org, another that can push to quay.io under my user).

So I had to work around this by doing the build and push ahead of time.

Maybe it's just my particular case was painful, so it's not that big of a deal.

gcs278 (commented):
Looks good, just a couple nit picks.

Makefile
alebedev87: > Is this new command just for troubleshooting?

Yes.

> Or will a downstream CI job be using this?

No, in the CI we will need to do something similar to [this](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/fips-check/openstack-k8s-operators-fips-check-commands.sh#L52).

Makefile
alebedev87: > Do you need to do build and push in this step?

Yes, `check-payload` needs the image to be pullable.

> I had a bit of pain with this, because to build and push I use different registry auth files (one that can pull the images from registry.ci.openshift.org, another that can push to quay.io under my user).

Interesting. I'm logged to my quay.io using `podman login`. The CI image registry auth is saved to `/run/user/${id}/containers/auth.json` thanks to `oc registry login` command. And they don't interfere on my machine.

gcs278: Yea this sounds like a personal problem I have with managing podman auth files. I should combine them.

Makefile
gcs278: > Yes, otherwise it gives the following error:

Hmm. If you can run the `image-build` and the `image-push` command with rootless podman, there's no reason you shouldn't be able to run this command with podman. My point being these should be consistent. 

Maybe the `$()` is causing issues, the other commands use `${}`? Try that:

```suggestion
	${CONTAINER_ENGINE} run --privileged $(CHECK_PAYLOAD_IMG) scan operator --spec $(IMG)
```

> In their examples they use sudo too. I didn't want to dig too deep to find out why.

They use `sudo` on the `./check-payload` command, not on the podman command.

It seems like in our other repos, we don't use sudo, like https://github.com/openshift/cluster-ingress-operator/blob/1217332a5a4e64b1ca333f817fa66bb6aae310ab/hack/release-local.sh#L26 or https://github.com/openshift/cluster-dns-operator/blob/afb52de2b4e18c2ed35f1bdb01f24a41a5efafc3/Makefile#L75.

> No, in the CI we will need to do something similar to this.

Ack.

alebedev87: > Hmm. If you can run the image-build and the image-push command with rootless podman, there's no reason you shouldn't be able to run this command with podman. My point being these should be consistent.

Agree about the consistency but `check-payload` is not a regular binary, it runs `podman` commands on its own ([podman pull](https://github.com/openshift/check-payload/blob/a0d1e0d35693c6d4c28f7f2e0f163e98dfdc67ae/internal/scan/scan.go#L283-L285) and others). So this turns into a "podman inside podman" kind of problem. May be there is a right combination of flags which would allow us to get rid of `sudo` but I didn't manage to find it quickly. I saw that `microshift` [uses `sudo` in their tests too](https://github.com/openshift/microshift/blame/6504bada8822cda998bd76420e0ccc9d36fdaed6/test/suites/fips/validate-fips.robot#L89) when they run `check-payload` as an image.

> Maybe the $() is causing issues, the other commands use ${}?

`$()` is the standard Makefile syntax, [equivalent](https://www.gnu.org/software/make/manual/html_node/Reference.html) to `${}`. I often use `$()` as it looks to me more "Makefile native".

> > In their examples they use sudo too. I didn't want to dig too deep to find out why.
> 
> They use sudo on the ./check-payload command, not on the podman command.

Yes, true. I was running `check-payload` built locally from the source code. Then I started looking for how to use their image to avoid referencing a third party tool in the make rule.

alebedev87:
`e2e-aws-rosa-operator` job is failing because the ROSA is experiencing [a problem](https://issues.redhat.com/browse/OSD-21663) with 4.16 release candidate.

Makefile
gcs278: > $() is the standard Makefile syntax, [equivalent](https://www.gnu.org/software/make/manual/html_node/Reference.html) to ${}. I often use $() as it looks to me more "Makefile native".

Ah yea sorry this isn't bash...Ack.

I think your laptop is having issues with the `podman run` command as rootless. Try out `podman run docker.io/hello-world` as an example. It looks like you can successfully run `podman build` and `podman push` with the other make commands. Might want to try a `podman system reset` to reset the storage to clear everything.

So you don't need `sudo` to run this container (or any container) with podman, works fine for me.

But it's a nit pick and I won't hold up the PR over it.

gcs278:
Thanks!
/unhold
/lgtm

Makefile
alebedev87: > So you don't need sudo to run this container (or any container) with podman, works fine for me.

Does this command works fine on your machine? 
```
podman run --privileged registry.ci.openshift.org/ci/check-payload:latest scan operator --spec quay.io/alebedev/albo:latest
```

gcs278: Yes sir, works fine for me.

```
$ podman run --privileged registry.ci.openshift.org/ci/check-payload:latest scan operator --spec quay.io/alebedev/albo:latest
I0328 22:18:36.752612       1 main.go:302] using embedded config
I0328 22:18:36.753011       1 types_config.go:12] using config &{Components:[] FailOnWarnings:false FilterFile: FromFile: FromURL: InsecurePull:false Limit:-1 ContainerImageComponent: ContainerImage: OutputFile: OutputFormat:table Parallelism:5 Java:false PrintExceptions:false PullSecret: TimeLimit:1h0m0s Verbose:false UseRPMScan:false ConfigFile:{FilterFiles:[] FilterDirs:[/lib/firmware /lib/modules /usr/lib/.build-id /usr/lib/firmware /usr/lib/grub /usr/lib/modules /usr/share/app-info /usr/share/doc /usr/share/fonts /usr/share/icons /usr/share/openshift /usr/src/plugins /rootfs /sysroot] FilterImages:[] JavaDisabledAlgorithms:[DH keySize < 2048 TLSv1.1 TLSv1 SSLv3 SSLv2 TLS_RSA_WITH_AES_256_CBC_SHA256 TLS_RSA_WITH_AES_256_CBC_SHA TLS_RSA_WITH_AES_128_CBC_SHA256 TLS_RSA_WITH_AES_128_CBC_SHA TLS_RSA_WITH_AES_256_GCM_SHA384 TLS_RSA_WITH_AES_128_GCM_SHA256 DHE_DSS RSA_EXPORT DHE_DSS_EXPORT DHE_RSA_EXPORT DH_DSS_EXPORT DH_RSA_EXPORT DH_anon ECDH_anon DH_RSA DH_DSS ECDH 3DES_EDE_CBC DES_CBC RC4_40 RC4_128 DES40_CBC RC2 HmacMD5] PayloadIgnores:map[openshift-enterprise-pod-container:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrNotDynLinked Files:[/usr/bin/pod] Dirs:[]}]} operator-lifecycle-manager-container:{FilterFiles:[/usr/bin/cpb /usr/bin/copy-content] FilterDirs:[] ErrIgnores:[]} ose-olm-rukpak-container:{FilterFiles:[/unpack] FilterDirs:[] ErrIgnores:[]}] TagIgnores:map[] RPMIgnores:map[containernetworking-plugins:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[] Dirs:[/usr/libexec/cni]}]} cri-o:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/bin/crio /usr/bin/crio-status] Dirs:[]} {Error:ErrNotDynLinked Files:[/usr/bin/pinns] Dirs:[]}]} cri-tools:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/bin/crictl] Dirs:[]}]} glibc:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrNotDynLinked Files:[/usr/sbin/ldconfig /sbin/ldconfig] Dirs:[]}]} glibc-common:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrNotDynLinked Files:[/usr/sbin/build-locale-archive] Dirs:[]}]} ignition:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/lib/dracut/modules.d/30ignition/ignition] Dirs:[]}]} podman:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/bin/podman /usr/libexec/podman/quadlet /usr/libexec/podman/rootlessport] Dirs:[]} {Error:ErrNotDynLinked Files:[/usr/libexec/podman/catatonit] Dirs:[]}]} podman-catatonit:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrNotDynLinked Files:[/usr/libexec/catatonit/catatonit] Dirs:[]}]} runc:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/bin/runc] Dirs:[]}]} skopeo:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrGoMissingTag Files:[/usr/bin/skopeo] Dirs:[]}]} tini:{FilterFiles:[] FilterDirs:[] ErrIgnores:[{Error:ErrNotDynLinked Files:[/usr/bin/tini-static] Dirs:[]}]}] ErrIgnores:[]}}
I0328 22:18:36.753076       1 main.go:102] "scan" version="0.3.1-91-ga0d1e0d3-dirty"
---- Successful run
```

Does `podman run docker.io/hello-world` work for you?

alebedev87: > Does podman run docker.io/hello-world work for you?

Yes.

I start to blame [this shortcoming](https://github.com/containers/podman/blob/main/rootless.md#shortcomings-of-rootless-podman) of rootless podman:

> Support for using native overlayfs as an unprivileged user is only available for Podman version >= 3.1 on a Linux kernel version >= 5.12

As I still use the RHEL CBS (kernel `4.18.0`). Mostly probably I'm the only one with the kernel that old. I have to migrate to the Fedora CBD anyway. So, I think that removing `sudo` makes sense for the future use indeed.

gcs278: That's odd you can run docker.io/hello-world fine. Maybe it's the `--privileged` flag causing issues.

PR #138, opened 2024-11-04
fix: Add AWS ISO partitions to LB CRD

gcampbell12:
ISO partitions are missing from `stsIAMRoleARN` meaning that the operator cannot be used in these partitions in Manual credentials mode without manually modifying the CRD. 

Partitions taken from https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.json


alebedev87:
Similar PR about the support of different partitions: https://github.com/openshift/aws-load-balancer-operator/pull/130.

alebedev87:
So far, we didn't have a request to support ISO partitions. The full support would require different IAM policies ([iso](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy_iso.json), [isob](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/iam_policy_isob.json)) from the upstream for non-STS clusters.

PR #180, opened 2025-07-03
NE-2064: Update base and builder images to UBI9

grzpiotrowski:
The RHEL ELS image had limited use and was previously chosen because UBI was not FIPS-ready. Now, UBI is the recommended base image for all customer-facing container images. It is also the most popular base image in Konflux.

UBI9 version was chosen to avoid fixable vulnerabilities of UBI8.

Changes to the work directories made due to container user permissions.


Thealisyed (commented):
/approve
/lgtm
/label docs-approved
/label px-approved
/label qe-approved

PR #187, opened 2025-07-16
NE-2064: Add runAsNonRoot to kube-rbac-proxy container

alebedev87:
This commit adds the `runAsNonRoot` field to the kube-rbac-proxy container in the operator deployment, preventing it from running as root and addressing the general Coverity warning: "The Kubernetes container is allowed to run as the root user."


Thealisyed (commented):
/lgtm
/approve

PR #195, opened 2025-09-04
Ensure tekton pipeline triggered only on source repository

grzpiotrowski:
Previously the tekton pipelines could be triggered from the events on a forked aws-load-balancer-operator repository if the Red Hat Konflux app was also enabled on that fork, for example for testing.
For the operator, this would cause a new image build from the fork and the bundle-hack/container_digest.sh file to be nudged in the source repository with the image digest from that fork's build.

This PR adds a check of the repository name in on-cel-expression to ensure that the pipelines are triggered only by events on the `openshift/aws-load-balancer-operator` repository.


grzpiotrowski:
/retitle NE-2063: Ensure tekton pipelines are triggered only on source repository

alebedev87 (approved):
/lgtm
/approve

PR #203, opened 2025-09-18
WIP: Add OLM Catalog for Konflux

grzpiotrowski:
WIP

Assuming all 1.2 references will be 1.3


catalog/aws-lb-optr-1-2/catalog.yaml
```diff
@@ -0,0 +1,88 @@
+---
+defaultChannel: stable-v1
+name: aws-load-balancer-operator
+schema: olm.package
+---
+entries:
+- name: aws-load-balancer-operator.v1.2.0
+name: stable-v1.2
+package: aws-load-balancer-operator
+schema: olm.channel
```
alebedev87, line 10: I forgot one thing. We have to define the upgrade graph in the channels. Which CSV the new bundle replaces and which ranges should be skipped. The latter is a quite a serious question, we used to skip all the versions which precede the new one to fast forward all the users to the latest version available. We should discuss which strategy we want to follow and whether the same one makes sense for both [`stable-vX` and stable-vX.Y` channels](https://github.com/openshift/aws-load-balancer-operator/blob/main/docs/versioning.md#channels).

bundle-hack/VERSION
alebedev87, line 1: Why do we need another version file?

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
```diff
@@ -11,11 +11,11 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "true"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
-    olm.skipRange: <1.2.0
+    olm.skipRange: <1.3.0
```
alebedev87, line 14: We need to check with OLM people whether we still need to set `olm.skipRange` taking into account that the same is set in FBC's channels. At least, we need to make sure this is not going to conflict with the FBC's contents.

Makefile
```diff
@@ -294,9 +294,17 @@ CATALOG_IMG ?= $(BUNDLE_TAG_BASE)-catalog:v$(BUNDLE_VERSION)
 # Directory for the file based catalog.
 CATALOG_DIR := catalog
 
+# Catalog version subdirectory based on BUNDLE_VERSION (used by Konflux)
+CATALOG_VERSION_DIR := aws-lb-optr-$(shell echo $(BUNDLE_VERSION) | sed 's/\([0-9]*\)\.\([0-9]*\)\..*/\1-\2/')
```
alebedev87, line 298: I thought that the convention is to use the package name inside `catalog/`. This will conflict with the existing package but can the new FBC be the only in the repository? It should be enough to enable the development, no?

catalog/catalog-template.yaml
```diff
@@ -0,0 +1,22 @@
...
+      mediatype: image/png
+    name: aws-load-balancer-operator
+    schema: olm.package
+  - entries:
+      - name: aws-load-balancer-operator.v1.3.0
+    name: stable-v1.3
```
alebedev87, line 10: No skipRange? Will the upgrade from stable-v1.2 channel work?

bundle/metadata/annotations.yaml
```diff
@@ -4,7 +4,7 @@ annotations:
   operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: aws-load-balancer-operator
-  operators.operatorframework.io.bundle.channels.v1: stable-v1.2,stable-v1
+  operators.operatorframework.io.bundle.channels.v1: stable-v1.3,stable-v1
```
alebedev87, line 7: Another thing to be checked with OLM.  Do we still need to add chanbels here? This is redundant with FBC channels.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
grzpiotrowski: As in the OLM docs ([link](https://olm.operatorframework.io/docs/concepts/olm-architecture/operator-catalog/creating-an-update-graph/#methods-for-specifying-updates)):

>In FBC, the upgrade edge metadata in bundles/CSVs is ignored, the source of truth for upgrade edges is the catalog itself.

bundle-hack/VERSION
grzpiotrowski: Oops, yes this is not needed. Accidental leftover from my testing.

config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml
alebedev87: Good, then we can keep the bundle manifests and metadata "as is" for the moment.

Makefile
grzpiotrowski: This versioned directory was an attempt at keeping the old non-konflux catalog in place, and having the catalog for Konflux at the same time. I was thinking the convention applied to the final catalog image. In this case this is handled by the containerfile, placing the generated catalog in `config/aws-load-balancer-operator` directory.

With the latest https://github.com/openshift/aws-load-balancer-operator/pull/203/commits/99f122e4399c54652d61eb78206b53f4158b9739 change I kept only the new catalog. Though this breaks the verify step as CI can't pull from the `redhat-users-workload` quay repository.
I'm thinking now I'll revert this last change and re-add the versioned catalog dir along the already existing one.

Makefile
```diff
@@ -295,13 +295,21 @@ CATALOG_IMG ?= $(BUNDLE_TAG_BASE)-catalog:v$(BUNDLE_VERSION)
...
+
 # Directory for the aws-load-balancer-operator package files.
 PACKAGE_DIR := $(CATALOG_DIR)/aws-load-balancer-operator
 
+.PHONY: generate-catalog
+generate-catalog: opm ## Generate catalog for the Konflux-built operator
```
alebedev87, line 305: To ensure that the Konflux FBC catalog is updated from the bundle image mentioned in template, we should add `generate-catalog` to `hack/verify-olm.sh`.

grzpiotrowski:
To validate the catalog I build the image locally and pushed to my quay.io.
Then created a CatalogSource CR in an Openshift cluster and checked the status of the CatalogSource, reporting READY state.

```bash
$ podman build -t quay.io/rh-ee-gpiotrow/aws-load-balancer-operator-catalog:v1.3.0 -f Containerfile.aws-load-balancer-operator-catalog .
```

```bash
$ podman push quay.io/rh-ee-gpiotrow/aws-load-balancer-operator-catalog:v1.3.0
```

```bash
$ oc get -n openshift-marketplace catalogsource fbc-test-catalogsource -oyaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  creationTimestamp: "2025-10-06T20:47:36Z"
  generation: 1
  name: fbc-test-catalogsource
  namespace: openshift-marketplace
  resourceVersion: "43191"
  uid: 9adb9b16-76df-4475-9c5a-1fc0aec0fc0e
spec:
  displayName: AWS Load Balancer Operator Custom Catalog
  image: quay.io/rh-ee-gpiotrow/aws-load-balancer-operator-catalog:v1.3.0
  publisher: Red Hat
  sourceType: grpc
status:
  connectionState:
    address: fbc-test-catalogsource.openshift-marketplace.svc:50051
    lastConnect: "2025-10-06T20:48:02Z"
    lastObservedState: READY
  registryService:
    createdAt: "2025-10-06T20:47:37Z"
    port: "50051"
    protocol: grpc
    serviceName: fbc-test-catalogsource
    serviceNamespace: openshift-marketplace
```

Makefile
grzpiotrowski: Done. I added that to the script.

PR #213, opened 2025-10-01
NE-2123: Bump bundle version to 1.3.0

alebedev87:
This PR prepares for the upcoming ALBO release by:
    - Updating the bundle version to `1.3.0`.
    - Adding a new minor OLM channel `1.3`.
    - Reading the bundle version from `VERSION` file.
    - Updating `make bundle` to run a `sed` command that sets the `skipRange` annotation to the current bundle version.

The last two changes ensure that bundle manifests remain in sync with the main source of truth for the version.


grzpiotrowski:
Thank you for the bump.
Should we also update the CSV at [config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml](https://github.com/openshift/aws-load-balancer-operator/blob/64f9280871423e809006d3374294efe2f774c28b/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml#L18)?

alebedev87:
> Should we also update the CSV at [config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml](https://github.com/openshift/aws-load-balancer-operator/blob/64f9280871423e809006d3374294efe2f774c28b/config/manifests/bases/aws-load-balancer-operator.clusterserviceversion.yaml#L18)?

I left it outdated. The bundle generation will replace the versions when `make bundle` (and later `make catalog`) will be run. The operator is supposed to be used from what's inside `bundle` (or `catalog`) directory. This makes 1 place fewer to take care about when a version is bumped. We can think of using placeholders in `config` directory, something to be tested.

PR #214, opened 2025-10-02
NE-2075: Set version on Konflux Containerfiles

alebedev87:
Add `make set-version` target to synchronize the `version` label in Konflux Containerfiles with the `VERSION` file (single source of truth).

Add `hack/verify-version.sh` to the `verify` target to ensure the version label is correctly set in Konflux Containerfiles.


Thealisyed (commented):
@alebedev87 For consistency across operators, should we add a README like we did in [external-dns-operator](https://github.com/Thealisyed/external-dns-operator/blob/43690d228b02db73416e18ce9c72ccd6020c5498/hack/README.md) for the sync-version.sh?

grzpiotrowski (approved):
That's great!

/lgtm
/approve

PR #218, opened 2025-10-07
NE-2123: Add LICENSE and regenerate Konflux FBC

alebedev87:
Add a LICENSE file to the Konflux catalog image to comply with the `ecosystem-cert-preflight-checks` Konflux check.

Regenerate the FBC catalog using the latest bundle built on Konflux to enable the `1.3` images.


alebedev87:
/override ci/prow/e2e-aws-proxy-operator
/override ci/prow/e2e-aws-rosa-operator
/override ci/prow/e2e-aws-operator
/override ci/prow/ci-bundle-aws-load-balancer-operator-bundle
/override ci/prow/ci-index-aws-load-balancer-operator-bundle-index
/override "Konflux kflux-prd-rh03 / aws-load-balancer-operator-container-aws-lb-optr-1-3-rhel-9-on-pull-request"

The PR changes only Konflux Containerfile and its contents. We don't even have a Konflux build pipeline for it. So it's really just a follow-up of https://github.com/openshift/aws-load-balancer-operator/pull/203.

alebedev87:
/override "kflux-prd-rh03 / aws-load-balancer-operator-container-aws-lb-optr-1-3-rhel-9-on-pull-request"

catalog/catalog-template.yaml
```diff
@@ -17,7 +17,7 @@ entries:
     skipRange: '<1.3.0'
     package: aws-load-balancer-operator
     schema: olm.channel
-  - image: quay.io/redhat-user-workloads/aws-load-balancer-operator-tenant/aws-lb-optr-1-3-rhel-9/aws-lb-bundle-container-aws-lb-optr-1-3-rhel-9@sha256:b98eb896c356c0b12a3dfa2f7aeee97c5a7eafd24c99dcafa01558677cbff6b0
+  - image: quay.io/redhat-user-workloads/aws-load-balancer-operator-tenant/aws-lb-optr-1-3-rhel-9/aws-lb-bundle-container-aws-lb-optr-1-3-rhel-9:881c3cdefcfc008395827b05cbd10808f5d8b5b2
```
grzpiotrowski, line 20: Previously we had the image referenced by the `@sha256:` digest instead of a tag.
This is because Konflux nudging mechanism does not work on tags, only on digests ([docs](https://konflux.pages.redhat.com/docs/users/building/component-nudges.html)).

But this applies more to the operator/operand images. In this case, we are always going to update the bundle reference manually, so using a tag makes more sense, as it is what we first see in the `Latest image` when going to the bundle component in the Konflux cluster.

grzpiotrowski (approved):
/lgtm
/approve

PR #221, opened 2025-10-08
NE-2123: Re-create FBC build pipelines with fbc-builder type

alebedev87:
After https://gitlab.cee.redhat.com/releng/konflux-release-data/-/merge_requests/11011 was synced into Konflux cluster. We have to regenerate FBC build pipelines with the new type: `fbc-builder`. This will make the FBC pipelines conformant to [Konflux best practices](https://konflux.pages.redhat.com/docs/users/end-to-end/building-olm.html#building-the-file-based-catalog):

> **Be sure to select the file-based catalog pipeline** and to specify the correct path to the catalog’s Containerfile within its git repository.


grzpiotrowski (approved):
/lgtm
/approve

PR #284, opened 2026-04-22
NE-2401: Add support for AWS ISO partitions

alebedev87:
Update IAM policy resource ARNs from `arn:aws:` to `arn:*:` to support all AWS partitions. Extend the `stsIAMRoleARN` CRD validation regex to accept `aws-iso`, `aws-iso-b`, `aws-iso-e`, and `aws-iso-f` partitions in addition to the existing `aws`, `aws-cn`, and `aws-us-gov`.

The CRD validation serves as the enforcement point for which partitions are supported, while `arn:*:` in the IAM policy resources ensures permissions work in whichever partition the cluster runs in. Note that the minified IAM policy used in non-STS clusters already collapses all resources to `*`, so the partition-specific ARNs only matter for the non-minified STS policy.


grzpiotrowski (approved):
/lgtm
/approve

PR #313, opened 2026-07-15
NE-2739: Migrate operator to native metrics protection with centralized TLS

bentito:
Migrates the `aws-load-balancer-operator` from utilizing `kube-rbac-proxy` sidecars to the native secure serving metrics server introduced in newer versions of `controller-runtime`.

### Key Changes:
- **`main.go`:** Integrated secure serving on port `:8443` using the native auth and authz filters. Added startup resolution of the OpenShift `APIServer` configuration named `cluster` to parse and apply the centralized `TLSSecurityProfile` (MinVersion, CipherSuites, and elliptic curves/groups, including post-quantum `X25519MLKEM768`). Restressed HTTP/2 by disabling it on the metrics server to match External DNS Operator (EDO) PR 502's security posture.
- **`config/`:** Deleted `kube-rbac-proxy` patches, updated `manager` deployment configuration to expose port `8443` directly, and commented out the auth proxy references in default Kustomizations.
- **`bundle/` & `vendor/`:** Regenerated the OLM bundle using `make bundle` and validated it successfully with `operator-sdk`. Resolved the vendored dependencies and updated the test CRD fixtures to keep test pipelines completely green.

Fully validated and verified locally using `make verify` and `make lint` with zero issues found.


go.mod
```diff
@@ -236,44 +253,61 @@ require (
...
+	golang.org/x/text v0.37.0 // indirect
+	golang.org/x/time v0.15.0 // indirect
+	golang.org/x/tools v0.44.0 // indirect
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
-	google.golang.org/protobuf v1.36.5 // indirect
-	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20250303144028-a0af3efb3deb // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250528174236-200df99c418a // indirect
+	google.golang.org/grpc v1.72.2 // indirect
```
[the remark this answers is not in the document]
bentito: These dependencies are managed centrally according to the team's release cadence. Upgrading them independently here is out of scope.

main.go
```diff
@@ -123,10 +125,28 @@ func main() {
...
 	})
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	profile := getTLSSecurityProfile(context.TODO(), restConfig)
+	tlsConfig, err := getTLSConfigFromProfile(profile)
+	if err != nil {
+		setupLog.Error(err, "unable to get TLS configuration from profile")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
bentito: The startup-only resolution is the standard team pattern for our operators (such as Ingress & EDO). Workload restarts handle any updates.

go.mod
```diff
@@ -236,44 +253,61 @@ require (
 	gitlab.com/bosi/decorder v0.4.2 // indirect
 	go-simpler.org/musttag v0.13.0 // indirect
 	go-simpler.org/sloglint v0.9.0 // indirect
+	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.61.0 // indirect
+	go.opentelemetry.io/otel v1.36.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.34.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.34.0 // indirect
+	go.opentelemetry.io/otel/metric v1.36.0 // indirect
+	go.opentelemetry.io/otel/sdk v1.36.0 // indirect
```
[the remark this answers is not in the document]
bentito: These dependencies are managed centrally according to the team's release cadence. Upgrading them independently here is out of scope.

bentito:
Unsupported TLS groups logging added and accidental files removed in commit 26970fc6.

hack/operator-permission-policy.json
```diff
@@ -2,26 +2,26 @@
...
     },
     {
-      "Action": [
+      "action": [
         "ec2:DescribeVpcs"
       ],
-      "Effect": "Allow",
-      "Resource": "*"
+      "effect": "Allow",
+      "resource": "*"
```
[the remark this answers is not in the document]
bentito: Added in commit 5750ba97

config/manifests/kustomization.yaml
```diff
@@ -19,7 +19,7 @@ patchesJson6902:
     # Remove the manager container's "cert" volumeMount, since OLM will create and mount a set of certs.
     # Update the indices in this path if adding or removing containers/volumeMounts in the manager's Deployment.
     - op: remove
-      path: /spec/template/spec/containers/1/volumeMounts/0
+      path: /spec/template/spec/containers/0/volumeMounts/0
```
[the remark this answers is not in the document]
bentito: Kustomize prepends the strategic-merged cert volumeMount to the beginning of the list, making it index 0. Re-running the generator with index 0 successfully removes the cert mount and preserves the sa-token and cabundle mounts perfectly (empty CSV diff).

bentito:
/assign @alebedev87 
This is pretty similar to your work on EDO https://github.com/openshift/external-dns-operator/pull/502 so hoping you could review this. Thanks!

bentito:
@davidesalerno I think the rosa test is a permafail until someone updates a role somewhere. Can you review when you get a chance please

alebedev87:
> I think the rosa test is a permafail until someone updates

Right, after we lost QEs some Prow CI steps were left unmaintained. We need to come back to this job and fix it. But for the moment we'd have to ignore it.

main.go
```diff
@@ -259,3 +279,88 @@ func getVPCId(ctx context.Context, ec2Client aws.EC2Client, clusterName string,
...
+
+	if len(profileSpec.Ciphers) > 0 {
+		ianaNames := libgocrypto.OpenSSLToIANACipherSuites(profileSpec.Ciphers)
+		var suites []uint16
+		for _, name := range ianaNames {
+			id, err := libgocrypto.CipherSuite(name)
+			if err != nil {
+				continue
+			}
```
alebedev87, line 330: We can add a log message similar to the unsupported group to leave a trace of unrecognized cipher suite.

main.go
```diff
@@ -259,3 +279,88 @@ func getVPCId(ctx context.Context, ec2Client aws.EC2Client, clusterName string,
...
+			suites = append(suites, id)
+		}
+		cfg.CipherSuites = suites
+	}
+
+	if len(profileSpec.MinTLSVersion) > 0 {
+		v, err := libgocrypto.TLSVersion(string(profileSpec.MinTLSVersion))
+		if err != nil {
+			return nil, fmt.Errorf("invalid TLS version %q: %w", profileSpec.MinTLSVersion, err)
+		}
```
alebedev87, line 340: How do we decide which of TLS settings are blocking like this minimal TLS version and which can be skipped like unrecognized ciphersuite of curves? Should we do the best effort here too and let the default (TLS1.2) be applied?

main.go
```diff
@@ -259,3 +279,88 @@ func getVPCId(ctx context.Context, ec2Client aws.EC2Client, clusterName string,
...
+			} else {
+				setupLog.Info("Skipping unsupported TLS group", "group", g)
+			}
+		}
+		if len(curves) > 0 {
+			cfg.CurvePreferences = curves
+		}
+	}
+
+	return libgocrypto.SecureTLSConfig(cfg), nil
```
alebedev87, line 358: `SecureTLSConfig` will fill unset settings with defaults. This makes the defaults we set in this function (e.g. min tls version) redundant. Should we just let `SecureTLSConfig` handle the unfilled defaults?

main.go
```diff
@@ -259,3 +279,88 @@ func getVPCId(ctx context.Context, ec2Client aws.EC2Client, clusterName string,
...
+	configv1.TLSGroupSecP256r1:      tls.CurveP256,
+	configv1.TLSGroupSecP384r1:      tls.CurveP384,
+	configv1.TLSGroupSecP521r1:      tls.CurveP521,
+	configv1.TLSGroupX25519MLKEM768: tls.X25519MLKEM768,
+}
+
+// TLSGroupToCurveID converts a configv1.TLSGroup name to its crypto/tls
+// CurveID. The second return value is false when the group is not supported
+// by the Go runtime.
+func TLSGroupToCurveID(group configv1.TLSGroup) (tls.CurveID, bool) {
```
alebedev87: Do we really need this function to be an exported symbol? As a matter of fact I don't see the reason for its existance, simple map lookup (`tlsGroupToCurveID[group]`) gives the same result.

main.go
```diff
@@ -259,3 +279,88 @@ func getVPCId(ctx context.Context, ec2Client aws.EC2Client, clusterName string,
...
+		cfg.MinVersion = v
+	}
+
+	if len(profileSpec.Groups) > 0 {
+		var curves []tls.CurveID
+		for _, g := range profileSpec.Groups {
+			if id, ok := TLSGroupToCurveID(g); ok {
+				curves = append(curves, id)
+			} else {
+				setupLog.Info("Skipping unsupported TLS group", "group", g)
```
alebedev87: To match other log messages:
```suggestion
				setupLog.Info("skipping unsupported TLS group", "group", g)
```

hack/generate-iam-from-credrequest.sh
```diff
@@ -15,4 +15,4 @@ cat <<EOF > "${POLICY_FILE}"
 }
 EOF
 ${YQ_BIN} -i -o=json "${POLICY_FILE}"
-sed -i -e 's/action/Action/g' -e 's/effect/Effect/g' -e 's/resource/Resource/g' "${POLICY_FILE}"
+python3 -c "import sys; p = sys.argv[1]; text = open(p).read().replace('\"action\"', '\"Action\"').replace('\"effect\"', '\"Effect\"').replace('\"resource\"', '\"Resource\"'); open(p, 'w').write(text)" "${POLICY_FILE}"
```
alebedev87: Is it just for a safer replacement? Or somehow related to the metrics change?

alebedev87 (commented):
Mostly LGTM. Some minor remarks and questions. One important thing I noticed though, [ALBO doesn't add servicemointor](https://github.com/openshift/aws-load-balancer-operator/blob/35a38c8aefb225c0e8444fc4734c3dced58747e0/config/default/kustomization.yaml#L24-L25) so the metrics are not actually scraped. This is gap which we can address as a separate PR.

main.go
bentito: Yep, added in 4ce1511b

main.go
bentito: The distinction comes down to fail-closed vs. fail-safe negotiation:
1. Fail-Closed for Protocol Versions: We cannot allow the metrics server to start up with a lower TLS protocol version than the global cluster policy specifies (e.g. falling back to TLS 1.2 when a strict TLS 1.3 profile is configured), as this would violate security compliance audits. Hence, an unrecognized protocol version must be a blocking startup error (fail-closed).
2. Fail-Safe for Ciphers/Curves: Unrecognized cipher suites and elliptic curves can be safely skipped (logged as warnings) because the TLS handshake protocol natively negotiates only the overlapping set of supported parameters. This is a fail-safe negotiation.

main.go
bentito: Agreed that it's technically redundant because `SecureTLSConfig` will populate the defaults. However, we explicitly initialize `MinVersion` inside the `tls.Config` declaration primarily to satisfy the gosec security linter (which flags any uninitialized or default TLS instantiation as G402: TLS MinVersion too low).

main.go
bentito: Agreed. Made the map unexported, removed the wrapper function in favor of a direct map lookup, and aligned the warning casing in commit 4ce1511b

main.go
bentito: Agreed. Aligned the log casing to match other messages in commit 4ce1511b

hack/generate-iam-from-credrequest.sh
bentito: The original `sed` command was platform-dependent and failed silently on macOS, which left uncommitted lowercase keys (action/effect/resource) in our local IAM permission policy compared to Linux CI. The Python replacement is cross-platform, allowing successful local schema generation and keeping the repository in sync

bentito:
> Mostly LGTM. Some minor remarks and questions. One important thing I noticed though, [ALBO doesn't add servicemointor](https://github.com/openshift/aws-load-balancer-operator/blob/35a38c8aefb225c0e8444fc4734c3dced58747e0/config/default/kustomization.yaml#L24-L25) so the metrics are not actually scraped. This is gap which we can address as a separate PR.

Agreed. Enabling the `ServiceMonitor` and wiring the Prometheus scraping permissions is a critical next step, but it is technically a separate task. I've tracked this follow-up work in the new Jira story:  https://redhat.atlassian.net/browse/NE-2840, so we can address it in a dedicated PR. Thanks!

hack/generate-iam-from-credrequest.sh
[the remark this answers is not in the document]
bentito: Agreed. Replaced the global text-replacement with a structured Python JSON parser to safely rename only the specific statement fields in commit 322397a2

hack/operator-permission-policy.json
```diff
@@ -24,4 +24,4 @@
       "Resource": "*"
     }
   ]
-}
+}
```
alebedev87: This change must be a result of `sed` replacement with `python`.

hack/generate-iam-from-credrequest.sh
alebedev87: https://github.com/openshift/aws-load-balancer-operator/pull/313#discussion_r3765778047

main.go
alebedev87: "unrecognized protocol version" should be a rare (or even impossible) scenario, sgtm.

hack/generate-iam-from-credrequest.sh
bentito: I have updated the generator script to explicitly append a trailing newline when writing out the JSON, and ran the generator to clean up the diff. Fixed in commit a72d83b

hack/operator-permission-policy.json
bentito: I have updated the generator script to explicitly append a trailing newline when writing out the JSON, and ran the generator to clean up the diff. Fixed in commit a72d83b

